PR #2, opened 2018-10-18
Fix out-of-bounds error within Gitea integration

tboerger:
Port from https://github.com/drone/drone/pull/2434

### DIFF
--- a/scm/driver/gitea/webhook.go
+++ b/scm/driver/gitea/webhook.go
@@ -190,27 +190,49 @@ func convertBranchHook(dst *createHook, action scm.Action) *scm.BranchHook {
 }
 
 func convertPushHook(dst *pushHook) *scm.PushHook {
-	return &scm.PushHook{
-		Ref: dst.Ref,
-		Commit: scm.Commit{
-			Sha:     dst.After,
-			Message: dst.Commits[0].Message,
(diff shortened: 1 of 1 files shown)


scm/driver/gitea/webhook.go
```diff
@@ -190,27 +190,41 @@ func convertBranchHook(dst *createHook, action scm.Action) *scm.BranchHook {
...
+			Repo:   *convertRepository(&dst.Repository),
+			Sender: *convertUser(&dst.Sender),
+		}
+	} else {
+		return &scm.PushHook{
+			Ref: dst.Ref,
+			Commit: scm.Commit{
+				Sha:       dst.After,
+				Link:      dst.Compare,
+				Author:    scm.Signature{},
```
bradrydzewski: maybe we use the `pusher` value here? https://github.com/drone/go-scm/blob/master/scm/driver/gitea/testdata/webhooks/push.json#L64:L70

tboerger: For author and commiter? Or only for one of them?

bradrydzewski: I'm thinking for both ...

tboerger:
I have updated it with the values from `Pusher`

PR #3, opened 2018-11-27
Paginate by URL

nathan-fps:
Fixes https://github.com/drone/drone/issues/2533. 

Allows for optionally paginating with a URL to next page instead of page numbers and modifies the Bitbucket driver repo list implementation to do this.

### DIFF
--- a/scm/driver/bitbucket/repo.go
+++ b/scm/driver/bitbucket/repo.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/drone/go-scm/scm"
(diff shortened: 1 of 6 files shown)


nathan-fps:
@bradrydzewski Can you give this a look when you have a chance?

scm/driver/bitbucket/repo.go
```diff
@@ -86,6 +87,9 @@ func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Pe
 // List returns the user repository list.
 func (s *repositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
 	path := fmt.Sprintf("2.0/repositories?%s", encodeListRoleOptions(opts))
+	if opts.URL != "" {
+		path = strings.TrimPrefix(opts.URL, s.client.BaseURL.String())
```
bradrydzewski: I think we can actually remove the `TrimPrefix` because when we try to combine the path with the base URL, if the path is a URL, it will return the URL. See https://play.golang.org/p/GXylQuPVrdx

So I think we can just simplify this to:

```
if opts.URL != "" {
		path = opts.URL
}
```

bradrydzewski:
thanks! sorry for the delayed review. 1 minor comment, but otherwise I think it looks great

scm/driver/bitbucket/repo.go
nathan-fps: Oh slick, I didn't release that was the behaviour

bradrydzewski: me neither actually ... I just wondered how it would behave if we passed a full URL and was pleasantly surprised by the results :)

nathan-fps: Ok, updated in the last commit

nathan-fps:
Not a problem! Thanks for the review

bradrydzewski:
ok, it may take a few days for me to get the dependency updated, but should be available early next week. Thanks for your help with this!

nathan-fps:
Great, looking forward to it!

PR #10, opened 2019-02-06
Fetch specific commit for Gitea

techknowlogick:
Hopefully this is suitable to enable cron tasks in Drone for Gitea repos


bradrydzewski:
this is great, thanks!

ogarcia:
I'm not GO developer but I think that same solution can be used to Gogs since the code is the same. The only difference is in line:
```
path := fmt.Sprintf("api/v1/repos/%s/git/commits/%s", repo, ref)
```

That in Gogs not have `/git/`:
```
path := fmt.Sprintf("api/v1/repos/%s/commits/%s", repo, ref)
```

Sample: https://try.gogs.io/api/v1/repos/gogs/gogs/commits/2c3e2b701e012294d457937e6bfbffd63dd8ae4f

Note that you must have authenticated to make this work:

```
curl -H "Authorization: token YOURTOKEN" https://try.gogs.io/api/v1/repos/gogs/gogs/commits/2c3e2b701e012294d457937e6bfbffd63dd8ae4f
```

PR #12, opened 2019-02-07
Fix: Problems with Yaml file being cutoff

NullP0interEx:
fixed https://discourse.drone.io/t/problems-with-yaml-file-being-cutoff/2941


bradrydzewski:
sorry, I did not see this and patched this myself a few weeks ago.

PR #24, opened 2019-05-15
Validate webhook using signature header in Gitea

techknowlogick:
WIP until I can run integration tests


techknowlogick:
@bradrydzewski Would this be possible to have a review before 1.2.0 is released?

bradrydzewski:
thanks!

PR #25, opened 2019-05-27
stash: replace login to slug for FindLogin

fitz123:
I didn't found any official description of slash's slug func, so This is pretty hack'y, but it makes the scm driver with logins like `name.lastname@company.name` at least usable

workaround for #31

### DIFF
--- a/scm/driver/stash/user.go
+++ b/scm/driver/stash/user.go
@@ -28,6 +28,7 @@ func (s *userService) Find(ctx context.Context) (*scm.User, *scm.Response, error
 	}
 	login := out.String()
 	login = strings.TrimSpace(login)
+	login = strings.Replace(login, "@", "_", -1)
 	return s.FindLogin(ctx, login)
 }
 


bradrydzewski:
I appreciate the pull request. I would prefer not to implement a hack or workaround and would rather make a permanent fix. Can you have help me understand the root cause of the problem and how we could permanently fix?

fitz123:
As you see here: https://github.com/drone/go-scm/pull/25/commits/f4d22d7fc6af4c65891374865890c5e523b1c057#diff-91862817fc90c211ea4d5904438f4310R30
_login_ is used to query user's account. But stash server expects slug instead. So _username@domain.com_ has 404 result meanwhile _username_domain.com_ would work.

I tried to map search request here #26 but I'm not so familiar with go but that could give you a clue of what going on

fitz123:
Ideally you instead of https://github.com/drone/go-scm/pull/25/files#diff-91862817fc90c211ea4d5904438f4310R31
one have to _slugify_ login with stash'es slug algorithm. But I didn't find any code or documentation of how stash'es slugify works besides used `@ -> _` replacement

gofort:
@bradrydzewski hi! I have the same problem in bitbucket ...  Can you please merge provided by @fitz123 fix?

bradrydzewski:
closing this in favor of the approach in #26

PR #26, opened 2019-05-31
stash: correct way to FindLogin using filter

fitz123:
alternative to pull #25


bradrydzewski:
Lets go with this approach. See comments in https://github.com/drone/go-scm/issues/31#issuecomment-523152286

bradrydzewski:
fixed by 0d1ea63283ade84c062be2b6ce5971ec9e1245cd

will be included in Drone 1.5.0 release

Issue #28, opened 2019-06-21
Bitbucket Author Username empty because of API changes


christianruhstaller:
Atlassian changed the API and it removed the "username" field from the user object. As documented here: 
https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-changes-gdpr/#removal-of-usernames-from-user-referencing-apis

This leads to an empty Username field and as a consequence the author field in drone will be empty. For example if you use the slack notification plugin the build.author field will be empty.

I think this issue has the same root: https://discourse.drone.io/t/branch-author-not-works-in-slack-plugin/4624

I think the a good replacement would be to use the `nickname` as a replacement for `username`.

This line is affected: https://github.com/drone/go-scm/blob/005219895eba392e5cd7ec66b94ac6a1c2f486e6/scm/driver/bitbucket/git.go#L147

christianruhstaller:
I added a pull request: https://github.com/drone/go-scm/pull/30

PR #30, opened 2019-06-21
Update Bitbucket driver to reflect API changes

christianruhstaller:
See Discussion: https://github.com/drone/go-scm/issues/28

### DIFF
--- a/scm/driver/gitlab/webhook.go
+++ b/scm/driver/gitlab/webhook.go
@@ -142,8 +142,9 @@ func convertPushHook(src *pushHook) *scm.PushHook {
 		},
 	}
 	if len(src.Commits) > 0 {
-		dst.Commit.Message = src.Commits[0].Message
-		dst.Commit.Link = src.Commits[0].URL
+		// get the last commit (most recent)
+		dst.Commit.Message = src.Commits[len(src.Commits)-1].Message
+		dst.Commit.Link = src.Commits[len(src.Commits)-1].URL
 	}
 	return dst
 }


scm/driver/bitbucket/git.go
```diff
@@ -207,14 +207,12 @@ func convertCommit(from *commit) *scm.Commit {
 			Name:   from.Author.User.DisplayName,
 			Email:  extractEmail(from.Author.Raw),
 			Date:   from.Date,
-			Login:  from.Author.User.Username,
```
tboerger: What happened to these fields?

christianruhstaller: Bitbucket changed their api and removed the "login" field:
https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-changes-gdpr/#removal-of-usernames-from-user-referencing-apis

tboerger:
Nickname is not a proper replacement as it doesn't have to be unique.

christianruhstaller:
If the Username needs to be unique. The Bitbucket API won't provide this information anymore. Down the road this means you cannot use the username anymore for usage as an information for plugins and co.

This breaks already: https://discourse.drone.io/t/branch-author-not-works-in-slack-plugin/4624

If the Username can't be used by the Bitbucket driver, drone needs to change its behavior to give the plugins a chance to use something like an username (Like the nickname or display name).

matthewhartstonge:
@tboerger if we want uniqueness, the best way forward would be to use the new Atlassian `account_id` field. Just did a poll of the API and got the following back for a commit:

```
"author": {
	"raw": "Matthew Hartstonge <matt@MY_DOMAIN.com>",
	"type": "author",
	"user": {
		"display_name": "Matthew Hartstonge",
		"uuid": "{A-COOL-DASH-SEPARATED-UUID}",
		"links": {
			"self": {
				"href": "https://api.bitbucket.org/2.0/users/%7BA-COOL-DASH-SEPARATED-UUID%7D"
			},
			"html": {
				"href": "https://bitbucket.org/%7BA-COOL-DASH-SEPARATED-UUID%7D/"
			},
			"avatar": {
				"href": "https://secure.gravatar.com/avatar/1414dfb6e23f3c4b2fab78cb84c3f070?d=https%3A%2F%2Favatar-management--avatars.us-west-2.prod.public.atl-paas.net%2Finitials%2FMH-4.png"
			}
		},
		"nickname": "Matthew Hartstonge",
		"type": "user",
		"account_id": "NUMBER:MY-SECR3T-ACCOUNT-ID"
	}
},

```

I'm happy to help out where I can, what would be the upstream implications/places it would need to be fixed?

tboerger:
That's the point where @bradrydzewski should jump in :)

bradrydzewski:
my thoughts are that the `username` field _must_ store a username. If Bitbucket no longer returns a username then this field will remain blank since no username is available. If we want to make another field available (e.g. nickname) we should add a new field.

matthewhartstonge:
@christianruhstaller you keen to continue with this PR?

christianruhstaller:
I removed the username field, as this does not exist in the Bitbucket API anymore and added the Nickname field which is not used currently.

PR #38, opened 2019-10-01
Added `CompareChanges` function to `GitService`.

chhsia0:
CompareChanges returns the changeset between two commits. If the source commit is not an ancestor of the target commit, it is up to the driver to return a 2-way or 3-way diff changeset.

### DIFF
--- a/scm/driver/bitbucket/git.go
+++ b/scm/driver/bitbucket/git.go
@@ -69,6 +69,14 @@ func (s *gitService) ListChanges(ctx context.Context, repo, ref string, opts scm
 	return convertDiffstats(out), res, err
 }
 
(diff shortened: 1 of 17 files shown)


chhsia0:
@bradrydzewski Would you mind taking a look at this PR? Thanks :)

bradrydzewski:
@chhsia0 thanks!  I really appreciate that you included extensive testing and support for as many providers as possible :)

tboerger:
As Gitea supports a diff view between branches I'm pretty sure there must also be an API for it. @lunny or @appleboy, maybe somebody of you could jump in with a followup pr?

lunny:
@tboerger will try this.

lunny:
I think we have an issue for that and need a PR. https://github.com/go-gitea/gitea/issues/5561

PR #40, opened 2019-10-09
fixes gitlab hook ssl verification

### LINKED ISSUE #39: Gitlab Hook sets incorrect value for parameter for skipping ssl verification
The Gitlab driver currently sets enable_ssl_verification to true when creating a hook if the input specifies to skip verification. It should set the paramater to false.

PR on its way.

ConorNevin:
Fixes #39


bradrydzewski:
thanks!  we will get this update included in the next Drone release

PR #41, opened 2019-10-23
Implemented `List` method for the content service.

### LINKED ISSUE #36: Support for listing the contents of a directory
**Feature request**
I would like to call `Client.Contents.Find` to discover which files and directories exist within a directory.

**Use case**
This is an issue for the implementation of drone config plugins that require traversing the source repository, such as `drone-tree-config`. See https://github.com/bitsbeats/drone-tree-config/issues/2

**Details**
The GitHub API call to `repos/%s/contents/path` accepts directories, but the source code of `go-scm` [assumes that you're calling it on a file](https://github.com/drone/go-scm/blob/master/scm/driver/github/content.go), causing it to crash because the response has a different format than the one expected. See https://developer.github.com/v3/repos/contents/#get-contents.

Is this a feature that you could see making it into `go-scm`, or should Drone plugins preferably implement their own git integration logic?

chhsia0:
List returns a list of contents in a repository directory by path. It is up to the driver to list the directory recursively or non-recursively, but a robust driver should return a non-recursive list if possible.

Currently only the `stash` driver returns a recursive list.

Fix #36.


scm/driver/gitlab/content.go
```diff
@@ -19,23 +19,12 @@ type contentService struct {
 }
 
 func (s *contentService) Find(ctx context.Context, repo, path, ref string) (*scm.Content, *scm.Response, error) {
-	path = url.QueryEscape(path)
-	path = strings.Replace(path, ".", "%2E", -1)
```
tboerger: This had been dropped without a replacement? Not sure if `url.QueryEscape` is really doing that?

scm/driver/gitlab/content.go
```diff
@@ -19,23 +19,12 @@ type contentService struct {
...
-	endpoint := fmt.Sprintf("api/v4/projects/%s/repository/files/%s?ref=%s", encode(repo), path, ref)
-	out := new(content)
+	endpoint := fmt.Sprintf("api/v4/projects/%s/repository/files/%s/raw?ref=%s", encode(repo), url.QueryEscape(path), ref)
+	out := new(bytes.Buffer)
 	res, err := s.client.do(ctx, "GET", endpoint, nil, out)
-	raw, berr := base64.StdEncoding.DecodeString(out.Content)
```
tboerger: Why has this been dropped?

scm/driver/gitlab/content.go
chhsia0: Oh yeah I should've added some more comments. The GitLab documentation says:
```
file_path (required) - Url encoded full path to new file. Ex. lib%2Fclass%2Erb
```
Which is arguably ambiguous about the meaning of "Url encoded," since dots are unreserved characters according to [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3).
And despite that the example in GitLab's documentation used `%2E`, the API server responded properly even if the dot is not replaced. So this replace is unnecessary.

That said, if you think it is safer to escape the dots, I'm happy to revert this change.

scm/driver/gitlab/content.go
chhsia0: Because I replaced the original API call (https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository) with another call that can directly return the raw data (see Line 22): https://docs.gitlab.com/ee/api/repository_files.html#get-raw-file-from-repository

scm/driver/gitlab/content.go
tboerger: I don't know how often this had been changed within the Gitlab API, as we are not only focussing on the hosted version it's possible that older Gitlab versions act differently. In the end it's up to brad if he's fine with dropping this encoding or if he wants to keep it as he will be hit by some possible support burden.

bradrydzewski: I would ask that we revert changes made to this method (both to encoding and changes to the endpoint).  Let's keep this pull request focused on `List` and we can discuss changes to this method in a separate pull request.

scm/driver/stash/git.go
```diff
@@ -77,37 +74,28 @@ func (s *gitService) ListCommits(ctx context.Context, repo string, opts scm.Comm
...
-	path := fmt.Sprintf("rest/api/1.0/projects/%s/repos/%s/tags?%s", namespace, name, encodeListOptions(opts))
+	endpoint := fmt.Sprintf("rest/api/1.0/projects/%s/repos/%s/tags?%s", namespace, name, encodeListOptions(opts))
 	out := new(branches)
-	res, err := s.client.do(ctx, "GET", path, nil, &out)
-	if !out.pagination.LastPage.Bool {
-		res.Page.First = 1
-		res.Page.Next = opts.Page + 1
-	}
+	res, err := s.client.do(ctx, "GET", endpoint, nil, out)
+	copyPagination(out.pagination, res)
```
bradrydzewski: can we move this change (and others in this file) to a separate pull request?  This will keep this PR isolated only to the List functionality.

bradrydzewski:
thanks!

the code looks excellent.  I do not have any major changes, but have request moving some code changes unrelated to `List` to a separate pull request. Note that variable name changes (e.g. to endpoint) are fine since these have low probability of regression.  Keeping the pull request isolated in scope just makes things a bit easier for us :)

chhsia0:
Sure I'll do that. Thanks :)

chhsia0:
@bradrydzewski Done PTAL!

PR #42, opened 2019-10-23
Cleaned up content service and added missing tests.

chhsia0:
The updated PR now consists of the following changes:
* Cleaned up some test code and added a missing test.
* Fixed Gitlab integration test.
* Used `url.PathEscape` instead of `url.QueryEscape` for url-encoding file paths in gitlab content service, as I suspect that the latter won't treat filenames with spaces correctly.


chhsia0:
If you guys think encoding dots is safer I could revert that part.

chhsia0:
Rebased on master so now the diff no longer includes changes in #41.

scm/driver/gitlab/content.go
```diff
@@ -20,23 +19,12 @@ type contentService struct {
...
 
 func (s *contentService) Find(ctx context.Context, repo, path, ref string) (*scm.Content, *scm.Response, error) {
-	path = url.QueryEscape(path)
-	path = strings.Replace(path, ".", "%2E", -1)
-	endpoint := fmt.Sprintf("api/v4/projects/%s/repository/files/%s?ref=%s", encode(repo), path, ref)
-	out := new(content)
+	endpoint := fmt.Sprintf("api/v4/projects/%s/repository/files/%s/raw?ref=%s", encode(repo), url.QueryEscape(path), ref)
+	out := new(bytes.Buffer)
 	res, err := s.client.do(ctx, "GET", endpoint, nil, out)
-	raw, berr := base64.StdEncoding.DecodeString(out.Content)
```
tboerger: That is something you should keep, that lead to issues in the past

bradrydzewski: agreed, we should keep this workaround in place. I honestly cannot remember why we needed it in the first place, but the last thing I want to do is change something and cause an outage for hundreds or even thousands of Drone installations.

bradrydzewski: edit: I see that this switches the API which does not encode the file, eliminating the base64 weirdness all together. Are there any concerns with file size? The base64 encoded API presumably limits file size to 1MB similar to GitHub which inherently protects Drone (primary consumer of this library) from downloading large files into memory. Maybe we should add an `io.LimitReader` here to mirror the 1MB limit in GitHub?

bradrydzewski:
> If you guys think encoding dots is safer I could revert that part.

yes, I think this would be best.  we know the current implementation works and I seem to recall issues with reverse proxies handling the encoding, but that was years ago so tough to say. We just cannot afford the risk of accidentally breaking something and impacting our customers.

bradrydzewski:
@chhsia0 could we move the following changes to a separate pull request:

scm/driver/gitlab/content.go
scm/driver/gitlab/content_test.go
scm/driver/gitlab/gitlab.go

I am good to merge all the formatting changes, however, I think we should have two separate functions (Find and FindRaw) for downloading the base64 ended file and the raw file similar to how these are handled in the google/go-github library:

- https://pkg.go.dev/github.com/google/go-github/v29/github?tab=doc#GitService.GetBlob
- https://pkg.go.dev/github.com/google/go-github/v29/github?tab=doc#GitService.GetBlobRaw

The reason I suggest two separate function calls is because in Drone we inherently rely upon the base64 endpoint and its 1mb limit to prevent a user from loading a huge file into the system. I would definitely merge a separate PR that breaks this out into a new function.

Thanks!

chhsia0:
@bradrydzewski Hi Brad. Sorry for keep delaying this. I withdrew the changes to use the raw file API in gitlab content service. Since I don't necessarily need to use the raw file API and it's not a generic API across different SCMs, I'm holding off on creating `FindRaw` until needed.

The updated PR now consists of the following changes:
* Cleaned up some test code and added a missing test.
* Fixed Gitlab integration test.
* Used `url.PathEscape` instead of `url.QueryEscape` for url-encoding file paths in gitlab content service, as I suspect that the latter won't treat filenames with spaces correctly.

PTAL :)

PR #47, opened 2019-11-21
Fix Bitbucket link handling for branches with slashes

christianruhstaller:
Bitbucket has a bug where the "source view" link for a branch
which contains a slash results in a 404.

The link to the "branch view" works with names containing
a slash so we do this for branches with slashes in its names.

See https://jira.atlassian.com/browse/BCLOUD-14422 for more information.

### DIFF
--- a/scm/driver/bitbucket/linker.go
+++ b/scm/driver/bitbucket/linker.go
@@ -24,8 +24,19 @@ func (l *linker) Resource(ctx context.Context, repo string, ref scm.Reference) (
 	case scm.IsPullRequest(ref.Path):
 		d := scm.ExtractPullRequest(ref.Path)
 		return fmt.Sprintf("%s%s/pull-requests/%d", l.base, repo, d), nil
-	case ref.Sha == "":
+	case scm.IsBranch(ref.Path) && ref.Sha == "":
 		t := scm.TrimRef(ref.Path)
+
+		// Bitbucket has a bug where the "source view" link for
+		// a branch which contains a slash results in a 404.
+		// The link to the "branch view" works with names containing
+		// a slash so we do this for branches with slashes in its names.
+		// See https://jira.atlassian.com/browse/BCLOUD-14422 for more information
+		//
+		if scm.BranchContainsSlash(t) {
+			return fmt.Sprintf("%s%s/branch/%s", l.base, repo, t), nil
+		}
+
 		return fmt.Sprintf("%s%s/src/%s", l.base, repo, t), nil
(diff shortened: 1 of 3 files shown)


scm/driver/bitbucket/linker.go
```diff
@@ -24,8 +24,19 @@ func (l *linker) Resource(ctx context.Context, repo string, ref scm.Reference) (
...
+	case scm.IsBranch(ref.Path) && ref.Sha == "":
 		t := scm.TrimRef(ref.Path)
+
+		// Bitbucket has a bug where the "source view" link for
+		// a branch which contains a slash results in a 404.
+		// The link to the "branch view" works with names containing
+		// a slash so we do this for branches with slashes in its names.
+		// See https://jira.atlassian.com/browse/BCLOUD-14422 for more information
+		//
+		if scm.BranchContainsSlash(t) {
```
bradrydzewski: since this function is only used by Bitbucket, it should be private an internal to the Bitbucket package, as opposed to a library-level function.

bradrydzewski: also, this does not really need to be a separate function, since it is just wrapping the container function with no extra logic.

```diff
-if scm.BranchContainsSlash(t) {
+if strings.Contains(t, "/") {
```

scm/driver/bitbucket/linker.go
```diff
@@ -24,8 +25,19 @@ func (l *linker) Resource(ctx context.Context, repo string, ref scm.Reference) (
...
+	case scm.IsBranch(ref.Path) && ref.Sha == "":
 		t := scm.TrimRef(ref.Path)
+
+		// Bitbucket has a bug where the "source view" link for
+		// a branch which contains a slash results in a 404.
+		// The link to the "branch view" works with names containing
+		// a slash so we do this for branches with slashes in its names.
+		// See https://jira.atlassian.com/browse/BCLOUD-14422 for more information
+		//
+		if strings.Contains(t, "/") {
```
bradrydzewski: one concern I have is that we could introduce a regression where the sha is empty, but the object type is not a branch. Could we adjust accordingly:

```diff
-case scm.IsBranch(ref.Path) && ref.Sha == "":
+ref.Sha == "":
```

```diff
-if strings.Contains(t, "/") {
+if scm.IsBranch(ref.Path) && strings.Contains(t, "/") {
```

christianruhstaller:
The suggestions are implemented, any more suggestions? Or is the current pull request good to merge?

Issue #48, opened 2019-11-25
Namespace in GitLab is wrong


odg0318:
Hello. I have an issue in GitLab repository.

Some repositories in my GitLab belong to sub-group, such as `foo/bar/baz/sample`.
In this case, `sample` is project name and `foo/bar/baz` is namespace in my opinion.
But actually only `baz` is parsed as namespace via this code.

Related codes are below.
https://github.com/drone/go-scm/blob/master/scm/driver/gitlab/repo.go#L172

As I know, `Drone` also uses this `drone/scm` project. And the above issue makes some errors to sync GitLab repository.

PR #49, opened 2019-11-25
`full_path` is used as namespace in GitLab repository.

odg0318:
I changed from `path` to `full_path` as namespace in  GitLab repository.

Related to https://github.com/drone/go-scm/issues/48


tboerger:
The namespaced paths will kill Drone as it's not prepared for that. AFAIK @bradrydzewski decided against applying this change as it's quite complex.

attie:
I have also just bumped in to #48 ... it seems that GitLab support is currently broken if you have a hierarchy deeper than 1.

@tboerger / @bradrydzewski do you know if this fix has been rejected?
If so, do you know if there are any plans to support GitLab properly?

attie:
Oh, I just found this (i.e: it's considered out of scope):

https://github.com/drone/drone/issues/2009#issuecomment-334540689

bradrydzewski:
while I appreciate the pull request we have decided we will not support subgroups in Drone. I am going to close this as out-of-scope accordingly.

PR #50, opened 2020-01-06
Fix gitea API call for triggering pipelines

qdeliens:
@bradrydzewski
Solution to the problem from [this issue.](https://discourse.drone.io/t/gitea-incorrect-api-call/6523)
All that remains is to change [drone/service/commit/commit.go line 90](https://github.com/drone/drone/blob/master/service/commit/commit.go#L90) to `commit, _, err := s.client.Git.FindCommitByRef(ctx, repo, ref)` to actually implement it.
Let me know if I need to make any changes regarding code style etc.

### DIFF
--- a/scm/driver/gitea/git.go
+++ b/scm/driver/gitea/git.go
@@ -25,19 +25,9 @@ func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Re
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
-	if strings.HasPrefix(ref, "refs/") {
-		path := fmt.Sprintf("api/v1/repos/%s/git/%s", repo, ref)
-		out := []*refInfo{}
-		_, err := s.client.do(ctx, "GET", path, nil, &out)
-		if err != nil {
-			return nil, nil, err
-		}
-		if len(out) == 0 {
-			return nil, nil, scm.ErrNotFound
-		}
-		ref = out[0].Object.Sha
-	}
-	path := fmt.Sprintf("api/v1/repos/%s/git/commits/%s", repo, ref)
+	ref = strings.TrimPrefix(ref, "refs/")
+	ref = strings.ReplaceAll(ref, "/", "%2F")
+	path := fmt.Sprintf("api/v1/repos/%s/commits/%s", repo, ref)
 	out := new(commitInfo)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
 	return convertCommitInfo(out), res, err
--- a/scm/driver/gitea/git.go
+++ b/scm/driver/gitea/git.go
@@ -90,20 +90,6 @@ type (
 		Committer user   `json:"committer"`
 	}
 
-	// gitea ref object.
-	ref struct {
-		Sha      string `json:"sha"`
-		Type     string `json:"type"`
-		URL      string `json:"url"`
-	}
-
-	// gitea ref info object.
-	refInfo struct {
-		Object   ref    `json:"object"`
-		Ref      string `json:"ref"`
-		URL      string `json:"url"`
-	}
-
 	// gitea signature object.
 	signature struct {
 		Name     string `json:"name"`
--- a/scm/driver/gitea/git.go
+++ b/scm/driver/gitea/git.go
@@ -30,6 +30,17 @@ func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Com
 	return convertCommitInfo(out), res, err
 }
 
+func (s *gitService) FindCommitByRef(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	path := fmt.Sprintf("api/v1/repos/%s/git/%s", repo, ref)
+	out := []*refInfo{}
+	_, err := s.client.do(ctx, "GET", path, nil, &out)
+    if err != nil {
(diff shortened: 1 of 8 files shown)


bradrydzewski:
Thanks for taking the time to send a pull request. 

Upon review I do not believe we need to add a new method for `FindCommitByRef`.  Instead we should be able to modify the existing `FindCommit for Gitea only:

```diff
FindCommit(ctx context.Context, repo, ref string) (*Commit, *Response, error) {
+	if strings.HasPrefix(ref, "refs/") {
+		// new API call
+	}

	path := fmt.Sprintf("api/v1/repos/%s/git/commits/%s", repo, ref)
	out := new(commitInfo)
	res, err := s.client.do(ctx, "GET", path, nil, out)
	return convertCommitInfo(out), res, err
}
```

scm/driver/gitea/git.go
```diff
@@ -30,6 +30,17 @@ func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Com
...
 }
 
+func (s *gitService) FindCommitByRef(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	path := fmt.Sprintf("api/v1/repos/%s/git/%s", repo, ref)
+	out := []*refInfo{}
+	_, err := s.client.do(ctx, "GET", path, nil, &out)
+    if err != nil {
+        return nil, nil, err
+    }
+    sha := out[0].Object.Sha
```
bradrydzewski: this code is subject to a panic if the index is out of bounds, so we should probably check the length of the slice just to be certain.

tboerger:
@appleboy @technowlogik @lunny does this make sense?

qdeliens:
> Thanks for taking the time to send a pull request.
> 
> Upon review I do not believe we need to add a new method for `FindCommitByRef`. Instead we should be able to modify the existing `FindCommit for Gitea only:
> 
> ```diff
> FindCommit(ctx context.Context, repo, ref string) (*Commit, *Response, error) {
> +	if strings.HasPrefix(ref, "refs/") {
> +		// new API call
> +	}
> 
> 	path := fmt.Sprintf("api/v1/repos/%s/git/commits/%s", repo, ref)
> 	out := new(commitInfo)
> 	res, err := s.client.do(ctx, "GET", path, nil, out)
> 	return convertCommitInfo(out), res, err
> }
> ```

This indeed makes more sense, I have pushed a new fix, also checking the slice length as you remarked.

lunny:
I think gitea could handle both commit id and refs, this PR is unnecessary. See 

https://github.com/go-gitea/gitea/blob/master/routers/api/v1/repo/commits.go#L56
https://github.com/go-gitea/gitea/blob/master/modules/git/repo_commit.go#L121

If you found any problem of gitea, please send an issue.

6543:
https://github.com/go-gitea/gitea/pull/10915 got merged ...

qdeliens:
> [go-gitea/gitea#10915](https://github.com/go-gitea/gitea/pull/10915) got merged ...

I have updated the code accordingly!

scm/driver/gitea/git.go
```diff
@@ -24,7 +25,11 @@ func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Re
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
-	path := fmt.Sprintf("api/v1/repos/%s/git/commits/%s", repo, ref)
+	ref = strings.TrimPrefix(ref, "refs/")
```
bradrydzewski, line 28: I noticed this code trims `refs/`.  Is it correct that `refs/heads/master` should be sent to the Gitea API as `heads/master`?  I just wanted to confirm ...

qdeliens: With the latest gitea image it was necessary yes. This is the new API call [go-gitea/gitea#10915](https://github.com/go-gitea/gitea/pull/10915). It would of course be nice if someone else could verify it also works for him/her.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: Hello @bradrydzewski , we spoke about this issue (failing cron builds on newer Gitea) in Drone's Gitter room last Friday, could I ask what is the current status? :)

CirnoT: `refs/heads/master` works fine, trimming is unnecessary.

lafriks:
API in Gitea has been changed back to not be breaking, so it should work again as it was but now for both commit sha hash and ref

CirnoT:
The changes made by this PR are now unnecessary. As for the original issue this PR tried to resolve (triggering pipeline via branch), the issue was never with Gitea API - it allows (and always did) passing ref or SHA commit.

We have however noticed potential reason why the problem exists - the `ref` should be URI component encoded, as in `refs/heads/master` should be sent as `refs%2Fheads%2Fmaster`.

qdeliens:
> The changes made by this PR are now unnecessary. As for the original issue this PR tried to resolve (triggering pipeline via branch), the issue was never with Gitea API - it allows (and always did) passing ref or SHA commit.
> 
> We have however noticed potential reason why the problem exists - the `ref` should be URI component encoded, as in `refs/heads/master` should be sent as `refs%2Fheads%2Fmaster`. I've opened new PR to handle that issue specifically #61.

#61 with the latest gitea build fixed the issue for me, closing this PR now.

PR #52, opened 2020-01-28
Fix gitea tag link

lunny:
As title.

### DIFF
--- a/scm/driver/gitea/linker.go
+++ b/scm/driver/gitea/linker.go
@@ -20,7 +20,7 @@ func (l *linker) Resource(ctx context.Context, repo string, ref scm.Reference) (
 	switch {
 	case scm.IsTag(ref.Path):
 		t := scm.TrimRef(ref.Path)
-		return fmt.Sprintf("%s%s/tag/%s", l.base, repo, t), nil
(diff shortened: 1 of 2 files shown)


lunny:
I think since 1.3.0, see https://github.com/go-gitea/gitea/commit/513375c429435ba60a667b219bdfb00e5b760b38

tboerger:
But gitea had some fallback mechanism if this part have not been added to the URL? Because this had been working so far.

jolheiser:
It looks like there was a redirect for "legacy" repositories in that commit, perhaps that worked for a while?  
The code for redirecting "legacy" is still in `master`, but it has changed a little.

lunny:
So it's now broken. See https://drone.gitea.com/gitea/go-sdk/254 .

tboerger:
> So it's now broken. See https://drone.gitea.com/gitea/go-sdk/254 .

I can't see where something should be broken...

lunny:
click the link on `v0.11.0`, you will get 404.

图片

tboerger:
You are right, I haven't realized that before. @bradrydzewski that fix should get into another release of Drone to fix the tag links for the Gitea integration.

PR #57, opened 2020-04-01
Change gitlab force_remove_source_branch to bool

jezhodges:
Gitlab merge requests currently give an error message:
`json: cannot unmarshal bool into Go struct field .object_attributes.merge_params.force_remove_source_branch of type string`

Gitlab appear to have changed `force_remove_source_branch` to a boolean in the API. This PR updates those structs.

### DIFF
--- a/scm/driver/gitlab/testdata/webhooks/pull_request_close.json
+++ b/scm/driver/gitlab/testdata/webhooks/pull_request_close.json
@@ -37,7 +37,7 @@
     "merge_commit_sha": null,
     "merge_error": null,
     "merge_params": {
-      "force_remove_source_branch": "0"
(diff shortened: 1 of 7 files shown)


bradrydzewski:
We should instead remove `MergeParams` from the structure since it is not being used.  The reason is that merging this change would fix Drone for the latest version of GitLab, but would also break Drone for anyone not using the latest version of GitLab.

jezhodges:
@bradrydzewski Good point. I've updated the structs to ignore MergeParams

jezhodges:
Any movement with this @bradrydzewski? As it's blocking a few of my pipeline jobs...

jezhodges:
@bradrydzewski any update on getting this merged?

PR #58, opened 2020-04-03
Set `Repository.Link` in various SCM drivers.

chhsia0:
The `Repository.Link` field can be useful in a number of places. For example, it can be used to construct the URL to the SCM web viewer of a given content file, especially when some major SCM services doesn't provide that in content service responses.


chhsia0:
@bradrydzewski Could you take a look when you have time? Thanks :)

bradrydzewski:
@chhsia0 thanks!

PR #59, opened 2020-04-08
Adds a test for when a GitHub deployment event ref is a SHA

mreifman:
I realize this isn't merge-able without a code fix, but I'm not sure how to approach the ref matching logic. This issue came up in https://discourse.drone.io/t/cloning-failed-in-github-deployment/6882/4 and @ashwilliams1 asked for a failure case.

Here's the failing test:

```
--- FAIL: TestWebhooks (0.02s)
    webhook_test.go:206: Error unmarshaling testdata/webhooks/deployment_commit.json
    webhook_test.go:207: {*scm.DeployHook}.Ref.Path:
        	-: "a10867b14bb761a232cd80139fbd4c0d33264240"
        	+: "refs/heads/a10867b14bb761a232cd80139fbd4c0d33264240"
```

How it fails depends on the SHA: if it starts with a digit then it matches the tag regex and gets `refs/tags` prepended.


bradrydzewski:
Thanks for providing the test case.  It is interesting that no branch or reference is provided but I guess it makes sense.  We will take it from here and patch the library to prevent branch and ref from being populated. We may find that this does not solve the problem, since I'm not sure how Drone will handle a build with no branch or reference, which means this may require additional patches to Drone core or the clone plugin.

mreifman:
For a bit more context, the build triggered by these deployment events in Drone 0.8 has a clone step that does:

```
git fetch --no-tags origin +refs/heads/master
...
git reset --hard -q ...
```

I had been thinking that the main thing to get this to work in 1.x was to get to `clone-commit` via the default case here: 

https://github.com/drone/drone-git/blob/1af95e77567b713bf66211955b1eedbfa2024c8c/posix/clone#L68-L70

But I'm seeing `DRONE_COMMIT_BRANCH` being set to the commit SHA, which won't work what `clone-commit` is currently doing:

https://github.com/drone/drone-git/blob/1af95e77567b713bf66211955b1eedbfa2024c8c/posix/posix_gen.go#L92

bradrydzewski:
Thanks for the details.  It looks like at a minimum we will also have to patch the clone plugin so it handles builds with no branch using a separate set of commands.

bradrydzewski:
fixed by https://github.com/drone/go-scm/commit/22d58836415f263e55e7a5e11005454ffd446dd2

PR #61, opened 2020-05-11
Ensure git ref is URI encoded (%2F) for Gitea call

CirnoT:
This fixes triggering pipeline via branch.

### DIFF
--- a/scm/driver/gitea/git.go
+++ b/scm/driver/gitea/git.go
@@ -7,8 +7,8 @@ package gitea
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"time"
-	"url"
 
 	"github.com/drone/go-scm/scm"
 )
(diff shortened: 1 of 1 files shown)


scm/driver/gitea/git.go
```diff
@@ -24,6 +25,7 @@ func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Re
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	ref = strings.Replace(ref, "/", "%2F", -1)
```
lafriks, line 28: imho it would be safer to use `url.QueryEscape`
```suggestion
	ref = url.QueryEscape(ref)
```

CirnoT: I think `PathEscape` should be used instead

lafriks: true :+1:

PR #62, opened 2020-05-25
fix FindCommit for branch name with slash for gitlab

hmldd:
fix FindCommit can not get commit for branch name with slash (like feature/new) from gitlab.
Reference: https://docs.gitlab.com/ee/api/commits.html#get-a-single-commit


tboerger:
@bradrydzewski should we just merge it?

bradrydzewski:
fixed by c29b3f1156cd6fb200364177aa6accc631e74082

PR #63, opened 2020-06-02
Supporetd `skip_cert_verification` for bitbucket webhook registration.

chhsia0:
This is undocumented in [Bitbucket REST API 2.0](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/hooks) but I tested it against `bitbucket.org` and it's a valid parameter. The UI also has a corresponding setting.


chhsia0:
@bradrydzewski Thanks for merging so quickly. Sorry I've been delaying on #42 but I'll find some time this week.

PR #66, opened 2020-06-19
Fix pagination being ignored for Gitea

### LINKED ISSUE #64: paginate Gitea repository list
Gitea implemented pagination for the repository list which results in only 10 repositories being returned. We have mechanisms in place for handling pagination. Here is a note from our Discourse forum:

> Pagination would need to be added to the code here:
https://github.com/drone/go-scm/blob/master/scm/driver/gitea/repo.go#L42
> 
> For reference you can see how we do this with GitHub:
https://github.com/drone/go-scm/blob/master/scm/river/github/repo.go#L85
> 
> Some additional code may be required to extract the pagination data from the response (i.e. current page, next page, last page, results per page). However, if Gitea uses the same approach as GitHub (example below) than this would not require any additional code.
> 
>```text
>Link: <https://api.github.com/user/repos?page=3&per_page=100>; rel="next",
>  <https://api.github.com/user/repos?page=50&per_page=100>; rel="last"
>```

_Source: https://discourse.drone.io/t/drone-only-shows-10-repositories-from-gitea/6922/16_

CirnoT:
Fixes #64

Gitea implementation was completely ignoring opts from Drone, so after we introduced proper pagination headers in v1.12, Drone would ask go-scm for page 2, and go-scm would request page 1 and return it, causing infinite loop of requests.

### DIFF
--- a/scm/driver/gitea/git.go
+++ b/scm/driver/gitea/git.go
@@ -36,8 +36,8 @@ func (s *gitService) FindTag(ctx context.Context, repo, name string) (*scm.Refer
 	return nil, nil, scm.ErrNotSupported
 }
 
-func (s *gitService) ListBranches(ctx context.Context, repo string, _ scm.ListOptions) ([]*scm.Reference, *scm.Response, error) {
-	path := fmt.Sprintf("api/v1/repos/%s/branches", repo)
+func (s *gitService) ListBranches(ctx context.Context, repo string, opts scm.ListOptions) ([]*scm.Reference, *scm.Response, error) {
+	path := fmt.Sprintf("api/v1/repos/%s/branches?%s", repo, encodeListOptions(opts))
 	out := []*branch{}
 	res, err := s.client.do(ctx, "GET", path, nil, &out)
 	return convertBranchList(out), res, err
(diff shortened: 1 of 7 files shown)


moepman:
Sorry for being slightly off-topic here, but my go skills are not good enough to tell it to use your patch when building drone. If you could give me a pointer how to achieve that I'd like to rebuild drone and report back if this actually fixes https://github.com/go-gitea/gitea/issues/11970

paulbsd:
Hi,

On your commit, there're mentions to "per_page" parameters but on the Gitea API version 1.12.0, per_page doesn't exists, and may be replaced by "limit".

Paul

CirnoT:
Ah sorry @paulbsd didn't notice that when copying util file from github implementation

CirnoT:
@moepman Before building drone you need to `go get` new version of go-scm.

CirnoT:
@bradrydzewski I've went ahead and added support for pagination on every endpoint. I do not guarantee that all of them support pagination right now, but If i am right then it won't matter anyway because Drone will never ask go-scm for next page if `Limit` header is not present.

moepman:
> @moepman Before building drone you need to `go get` new version of go-scm.

I still didn't quite figure out how to tell go get use your branch instead of upstream...
However: putting the contents of `CrinoT/patch-1` inside `go/pkg/mod/github.com/drone/go-scm@v1.6.1-0.20200417160414-b3466885f96` and then running `go install github.com/drone/drone/cmd/drone-server` resulted in a drone-server binary that would correctly display all of my 33 repositories from my gitea 1.12 installation :)
So I can confirm that this should fix the issue.

PR #67, opened 2020-07-17
Support fetching labels from pull requests in github,gitlab & gitea

takirala:


bradrydzewski:
this looks good.  my only request would be to ensure that one of the sample pull request webhooks for GitHub has labels so that the conversion is covered by unit tests.

bradrydzewski:
on second thought ... I wonder if we should return an array of `Label` structures instead of a string array.  The downside to a string array is that if we ever decide we need to expose more details, it would require a breaking change. So perhaps it is just best to provide the full details from the onset ...

```
Label struct {
  Name string
  Color string
}
```

takirala:
@bradrydzewski i have added 

```
Label struct {
  Name  string
}
```

i did not add color field for this reason : gitlab does not provide it in merge api response. it is however "fetchable" by making another API call. also, since this change would be now backwards compatible we can always add a color field (or any other fields) in the future as needed. 

Also updated the unit tests to validate the conversions. lmk what you think.

bradrydzewski:
I plan on adding methods to create / find / list / delete labels so it would be safe to add the `Color` field in the structure at this time :)  It is ok that the field would not be populated by GitLab.  We make a best effort to populate all fields, but there are a number of instances where some webhooks have empty fields for values that are less critical.

takirala:
@bradrydzewski sounds good, updated the pr!

bradrydzewski:
thanks!

PR #74, opened 2020-08-26
Add support for commit list in push webhook

shubhag:
This patch adds support for commits in push web hook for GitHub, bitbucket, gitlab, gogs, stash and gitea


bradrydzewski:
Thanks, this looks great!  The only feedback I have is that we should add the mappings to the Stash and Gitea drivers as well.  It is important that we try to maintain feature parity across all drivers as much as possible :)

https://github.com/drone/go-scm/blob/master/scm/driver/gitea/testdata/webhooks/push.json#L6:L32
https://github.com/drone/go-scm/blob/master/scm/driver/stash/testdata/webhooks/push.json#L30:L42

shubhag:
Added support for gitea, gogs and stash

PR #75, opened 2020-09-02
Fix github pull request link from diff url to html url

shubhag:


tboerger:
Such changes got to be applied to all supported drivers.

shubham149:
Added gitea support for diff URL

shubham149:
Added Github, bitbucket & gitea support for diff url. For others, diff url is not present in pr.json. Also, only github was using diff url instead of html url for link.

PR #78, opened 2020-10-09
Add github ping hook handler

rishabh-arya95:
When creating a new webhook GitHub sends the ping event to see if the setup is done properly. Currently, the ping is not handled, and on receiving the ping event `Unknown webhook event`.

### DIFF
--- a/scm/driver/github/webhook.go
+++ b/scm/driver/github/webhook.go
@@ -442,7 +442,6 @@ func convertDeploymentHook(src *deploymentHook) *scm.DeployHook {
 
 func convertPingHook(src *pingHook) *scm.PingHook {
 	return &scm.PingHook{
-		// Action        Action
 		Repo: scm.Repository{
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Thanks.  In order to merge this pull request, it needs to include an implementation for bitbucket, gitlab and gitea. This ensures the implementation and data model are compatible across providers, and avoids a situation where this library has github-specific implementations.

rishabh-arya95:
Thanks for your reply @bradrydzewski, I am primarily a GitHub user so did not look into others, but your point is valid. I will try to implement similar functionality for other providers as well.

eoinmcafee00:
hey @rishabh-arya95 again thank you for submitting this PR. It's been a while since there's been any activity and unless you are planning on implementing the full functionality as Brad commented above, we are going to have to close this one.

Let me know your thoughts.

Cheers,
Eoin

Issue #80, opened 2020-10-28
Adding code for files changed in the commit range?


jimsheldon:
This was brought up in https://github.com/meltwater/drone-convert-pathschanged/issues/34

Would it make sense to add functions to this project similar to https://github.com/meltwater/drone-convert-pathschanged/blob/master/providers/github.go#L13 which can determine the files changes in a commit range?

If so, we can work on a pull request.

bradrydzewski:
I think we may have all functions needed.  We have both `FindCommit` and a `CompareChanges` function:
https://github.com/drone/go-scm/blob/master/scm/driver/github/git.go#L26
https://github.com/drone/go-scm/blob/master/scm/driver/github/git.go#L68

jimsheldon:
Thanks, I will put something in our backlog around this, we'll get a pull request together soon.

PR #81, opened 2020-10-30
add Files to scm.Commit

jimsheldon:
the list of files changed will be useful when taking action based on files
changed

this will hopefully handle https://github.com/drone/go-scm/issues/80 and let https://github.com/meltwater/drone-convert-pathschanged use this module rather than its own logic

it is unclear to me if changing `scm.Commit` will affect other non-github drivers, but since `FindCommit` calls `convertCommit`, which returns `*scm.Commit`, I can't see a way around it

### DIFF
--- a/scm/driver/github/git.go
+++ b/scm/driver/github/git.go
@@ -126,10 +126,20 @@ func convertCommitList(from []*commit) []*scm.Commit {
 }
 
 func convertCommit(from *commit) *scm.Commit {
+	var files []scm.File
+	for _, file := range from.Files {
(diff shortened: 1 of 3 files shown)


bradrydzewski:
hey Jim, are you able to use the [ListChanges](https://github.com/drone/go-scm/blob/master/scm/git.go#L70) function?  This function is provider-agnostic and works for GitHub, Bitbucket, Bitbucket Server and GitLab.

jimsheldon:
`ListChanges` looks perfect, wish I'd seen that sooner, thanks!

PR #83, opened 2020-11-07
Add a vet step to drone config

tboerger:


eoinmcafee00:
Thanks @tboerger . Merging. now.

PR #84, opened 2020-11-18
fix(bitbucket) : Modify PushHook parse

beaucoralk:
* Add "After" to matche the "change.fromHash" from the recent Bitbucket payload push event (fast fix for the skip-ci issue)

linked to : https://github.com/drone/drone/pull/3038

### DIFF
--- a/scm/driver/bitbucket/webhook.go
+++ b/scm/driver/bitbucket/webhook.go
@@ -124,7 +124,7 @@ type (
 		Push struct {
 			Changes []struct {
 				Forced bool `json:"forced"`
-				FromHash string `json:fromHash`
+				ToHash string `json:toHash`
 				Old    struct {
 					Type  string `json:"type"`
 					Name  string `json:"name"`
@@ -416,7 +416,7 @@ func convertPushHook(src *pushHook) *scm.PushHook {
 	namespace, name := scm.Split(src.Repository.FullName)
 	dst := &scm.PushHook{
 		Ref:    scm.ExpandRef(change.New.Name, "refs/heads/"),
-		After:  change.FromHash,
+		After:  change.ToHash,
 		Before: change.Old.Target.Hash,
 		Commit: scm.Commit{
 			Sha:     change.New.Target.Hash,
--- a/scm/driver/bitbucket/webhook.go
+++ b/scm/driver/bitbucket/webhook.go
@@ -124,6 +124,7 @@ type (
 		Push struct {
 			Changes []struct {
 				Forced bool `json:"forced"`
+				FromHash string `json:fromHash`
 				Old    struct {
 					Type  string `json:"type"`
 					Name  string `json:"name"`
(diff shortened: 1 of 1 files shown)


beaucoralk:
I've not realy tested it, but the goal is to add the "After" commit hash and with https://github.com/drone/drone/pull/3038, the message could be filled.

bradrydzewski:
the `After` field should store the SHA of the most recent commit on ref after the push.  Is the `FromHash` the correct value to populate here?  Also the sample payloads in our unit tests do not include a `fromHash` value.  Where does this come from?

bradrydzewski:
I noticed you are trying to add `toHash` and `fromHash` fields.  These fields do not exist in the Bitbucket Cloud implementation, but they do exist in the Bitbucket Server (Stash) implementation.  Perhaps you mean to be looking at the code in this directory instead? https://github.com/drone/go-scm/tree/master/scm/driver/stash

Note that we recently (just a few days ago) merged a pull request that sets Before and After fields for Stash (Bitbucket Server). See https://github.com/drone/go-scm/pull/82.

beaucoralk:
> 
> 
> the `After` field should store the SHA of the most recent commit on ref after the push. Is the `FromHash` the correct value to populate here? Also the [sample payloads](https://github.com/drone/go-scm/blob/master/scm/driver/bitbucket/testdata/webhooks/push.json) in our unit tests do not include a `fromHash` value. Where does this come from?

Yeah sorry, I've changed to use the "ToHash". I've actualy the last self-hosted Bitbucket server installed in my Kube, and I found this in the official doc : https://confluence.atlassian.com/bitbucketserver/event-payload-938025882.html

bradrydzewski:
not sure if you saw my previous comment, but the code in these files is for Bitbucket Cloud.  The code for Bitbucket Server (Stash) can be found at [github.com/drone/go-scm/scm/driver/stash](https://github.com/drone/go-scm/tree/master/scm/driver/stash).  We also very recently merged pull request [#82](https://github.com/drone/go-scm/pull/82) which sets the Before and After values for Stash.

beaucoralk:
> 
> 
> I noticed you are trying to add `toHash` and `fromHash` fields. These fields do not exist in the Bitbucket Cloud implementation, but they do exist in the Bitbucket Server (Stash) implementation. Perhaps you mean to be looking at the code in this directory instead? https://github.com/drone/go-scm/tree/master/scm/driver/stash
> 
> Note that we recently (just a few days ago) merged a pull request that sets Before and After fields for Stash (Bitbucket Server). See #82.

Yeah, sorry, I've not well look up for. Then you should see the pull request about change the order fo fill the message before do the skip trigger.

Is it in the 1.9.2 release of drone ?

bradrydzewski:
I will take a look at the other pull request ...  let's move the discussion there.  I am going to close this PR since Bitbucket Server has the Before and After fields populated.  The 1.9.2 release does not have the latest version of the go-scm module, which means the dependency needs to be updated to take advantage of these features. Cheers.

PR #89, opened 2020-12-30
use access json header / extend error message parsing for stash

bakito:
As found in https://discourse.drone.io/t/panic-or-undefined-error-when-create-build-from-bitbucket-repository/8549/3

Somehow Bitbucket server does not always return a list of Errors, but only one message with status code.

This PR uses this rool-level message as error if present if the error list is empty.

It also sets the "Accept: application/json" header on the request, since the following code expects the response to be json.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@
 *.out
 _*.md
 .vscode
+.idea
 cmd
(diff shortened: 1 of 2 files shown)


bakito:
@ashwilliams1  sorry had the wrong branch in the PR, it's now corrected and all dev code is removed

bradrydzewski (approved):

bakito:
When will this be merged and integrated into drone?

eoinmcafee00:
Thanks @bakito 

Looks good merging now.

PR #90, opened 2020-12-30
retry with event subset for legacy stash versions

bakito:
stash 7.x supports more hook events that 6.x therefore the event setup should be possible by defining the stash version.

This PR implements the solution proposed here: https://discourse.drone.io/t/attempt-to-enable-repo-on-older-bitbucket-servers-fails/8486

The client is intentionally kept backwards compatible, to not break the drone server when go-scm is updated.

### DIFF
--- a/scm/driver/stash/repo.go
+++ b/scm/driver/stash/repo.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/drone/go-scm/scm"
 )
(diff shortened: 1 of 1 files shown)


bradrydzewski:
hey there, I appreciate the pull request.  I think we can simplify this a bit without requiring the user provide their bitbucket server version.  Also I'm not too keen to add a third party runtime dependency;  this library currently does not have any runtime dependencies outside of the Go standard library.

I think the easiest solution would be to detect the error in legacy versions of the software and then retry:

```
	res, err := s.client.do(ctx, "POST", path, in, out)
+	if err != nil && isUnknownEvent(err) {
+		removeUnknownEvent(in)
+		res, err = s.client.do(ctx, "POST", path, in, out)
+	}
	return convertHook(out), res, err
```

```
+func isUnknownEvent(err error) bool {
+	return strings.Contains(err, "pr:from_ref_updated is unknown")
+}
+
+func removeUnknownEvent(in *hookInput) {
+	var events []string
+	for _, event := range in.Events {
+		if event != "pr:from_ref_updated" {
+			events = append(events, event)
+		} 
+	}
+	in.Events = events
+	return in
+}
```

bakito:
I could successfully verify the simplified implementation in my setup.

madhosoi:
Any schedule to have this fix deployed on version 2 or latest?

PR #92, opened 2021-01-07
GitHub: add "Created" field into pushHook

quantonganh:
In GitHub, when pushing to a new branch, two hooks are created:
- X-GitHub-Event: create
- X-GitHub-Event: push

Since Drone ignores the "create" event, we needs a way to know that when a new branch has been created in "push" event.
This PR is created to do that.

### DIFF
--- a/scm/driver/github/testdata/webhooks/push_branch_create.json.golden
+++ b/scm/driver/github/testdata/webhooks/push_branch_create.json.golden
@@ -40,5 +40,6 @@
     "Name": "",
     "Email": "",
     "Avatar": "https://avatars1.githubusercontent.com/u/817538?v=4"
-  }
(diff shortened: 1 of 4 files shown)


bradrydzewski:
:wave: hey there, we actually don't need to add a new flag to know if the branch was created.  Instead we can check for an [empty commit](https://github.com/drone/go-scm/blob/cb9eef77a93f78f54ce03bc61faf462f742907cc/scm/driver/github/testdata/webhooks/push_branch_create.json.golden#L4) like this:

```
PushHook.Before == scm.EmptyCommit
```

quantonganh:
You're right. Let me update the PR https://github.com/drone/drone/pull/3052.

PR #93, opened 2021-01-10
[CI-0]: Added Pr in issue

aman-harness:
Currently, there is no way to verify if a webhook coming from github is a comment on PR or a comment on an issue. The difference b/w the webhooks is that the latter one has `pull_request` key in it. We need to forward the info to scm.IssueCommentHook so that the downstram service can understand if which type of comment this is.

### DIFF
--- a/scm/driver/github/issue.go
+++ b/scm/driver/github/issue.go
@@ -91,13 +91,14 @@ func (s *issueService) Unlock(ctx context.Context, repo string, number int) (*sc
 }
 
 type issue struct {
-	ID      int    `json:"id"`
-	HTMLURL string `json:"html_url"`
-	Number  int    `json:"number"`
-	State   string `json:"state"`
-	Title   string `json:"title"`
-	Body    string `json:"body"`
-	User    struct {
+	ID          int    `json:"id"`
+	HTMLURL     string `json:"html_url"`
+	Number      int    `json:"number"`
+	State       string `json:"state"`
+	Title       string `json:"title"`
+	Body        string `json:"body"`
+	PullRequest pr     `json:"pull_request"`
+	User        struct {
 		Login     string `json:"login"`
 		AvatarURL string `json:"avatar_url"`
 	} `json:"user"`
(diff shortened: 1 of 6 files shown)


scm/driver/github/testdata/issue.json.golden
```diff
@@ -15,5 +15,9 @@
         "Avatar": "https://github.com/images/error/octocat_happy.gif"
     },
     "Created": "2011-04-22T13:33:48Z",
-    "Updated": "2011-04-22T13:33:48Z"
+    "Updated": "2011-04-22T13:33:48Z",
+	"PullRequest": {
```
tboerger, line 19: Looks like a different indentation.

scm/driver/github/testdata/webhooks/comment.json.golden
```diff
@@ -1,58 +1,62 @@
...
-	},
-	"Sender": {
-		"Login": "lts-def",
-		"Name": "",
-		"Email": "",
-		"Avatar": "https://avatars0.githubusercontent.com/u/10278482?v=4",
-		"Created": "0001-01-01T00:00:00Z",
-		"Updated": "0001-01-01T00:00:00Z"
-	}
+   "Action":"created",
```
tboerger, line 2: Changing the indentation makes it much harder to see the changes.

scm/driver/github/issue.go
```diff
@@ -91,13 +91,14 @@ func (s *issueService) Unlock(ctx context.Context, repo string, number int) (*sc
...
-	Title   string `json:"title"`
-	Body    string `json:"body"`
-	User    struct {
+	ID          int    `json:"id"`
+	HTMLURL     string `json:"html_url"`
+	Number      int    `json:"number"`
+	State       string `json:"state"`
+	Title       string `json:"title"`
+	Body        string `json:"body"`
+	PullRequest pr     `json:"pull_request"`
```
tboerger, line 100: The attribute will container diff and link only? If you use the `pr` type it will contain a whole empty struct just filled by the two values.

Issue #76, opened 2020-09-20
GitHub CommitListOptions Parameter is Out of Date


blueworrybear:
It seems to me that one of `Git.ListCommits` arguments `CommitListOptions` requires updating.
It uses `ref` to query commits from branch, but the GitHub official documentation ([List Commits Parameters](https://docs.github.com/en/rest/reference/repos#list-commits--parameters)) says the query parameter should be `sha`

```go
func encodeCommitListOptions(opts scm.CommitListOptions) string {
	params := url.Values{}
	if opts.Page != 0 {
		params.Set("page", strconv.Itoa(opts.Page))
	}
	if opts.Size != 0 {
		params.Set("per_page", strconv.Itoa(opts.Size))
	}
	if opts.Ref != "" {
		params.Set("ref", opts.Ref)
	}
	return params.Encode()
}
```

Using `ref` to query branch commits will always return commits from default (master) branch.
Please let me know if I am wrong. I could also create PR to update it if needed.

PR #94, opened 2021-01-21
Update GH list-commits parameter

Yshayy:
Fixes: #76 

From local testing against GH api, it seems this change is still compatible with passing ref in the sha parameter (for example, branch name)


Yshayy:
Hi :)
Is there any missing code/context that I should add?

PR #95, opened 2021-02-10
Add webhook support for codecommit

shubham149:
AWS codecommit triggers handles
1. push to existing branch
2. branch create/delete
3. tag create/delete


scm/driver/codecommit/testdata/webhooks/branch_create.json
```diff
@@ -0,0 +1,12 @@
+{
```
shubham149, line 1: @bradrydzewski I wanted to check whether branch create webhook will come under branch hook or push hook. Codecommit does not send both push & branch create trigger when a branch is created.

PR #100, opened 2021-03-25
(feat) bitbucket cloud returns hash in Content.find

tphoney:


scm/driver/bitbucket/content.go
```diff
@@ -20,10 +20,18 @@ func (s *contentService) Find(ctx context.Context, repo, path, ref string) (*scm
...
 		Path: path,
 		Data: out.Bytes(),
-	}, res, err
+	}
+	if err != nil {
+		return content, res, err
+	}
+	metaEndpoint := fmt.Sprintf("/2.0/repositories/%s/src/%s/%s?format=meta", repo, ref, path)
+	metaOut := new(metaContent)
+	metaRes, err := s.client.do(ctx, "GET", metaEndpoint, nil, metaOut)
```
bradrydzewski, line 32: have your confirmed meta API call works for pull request and tag refs?  Also, I was wondering if a failure to fetch the meta should result in an error or not.  Is there any risk that returning the file may succeed but returning the meta may fail ?  This is me just being extra cautious because Drone relies on this function, but we would not want a meta failure from preventing a pipeline from running.

scm/driver/bitbucket/content.go
```diff
@@ -57,6 +65,13 @@ type content struct {
 	Attributes []string `json:"attributes"`
 }
 
+type metaContent struct {
+	Path   string `json:"path"`
+	Commit *struct {
```
bradrydzewski, line 70: the pointer implies the value can be null, but at line 33 we do not do a null check.  Do we need to worry about a null pointer here?  I would recommend we remove the pointer mostly for the purpose of consistency with the rest of the code.

bradrydzewski: FYI this is a really useful tool for generating structs from JSON samples https://mholt.github.io/json-to-go/.  Less useful for smaller JSON objects like this one, but very useful for larger objects.

PR #103, opened 2021-04-12
(feat) add create branch functionality

tphoney:


scm/git.go
```diff
@@ -13,6 +13,12 @@ import (
 const EmptyCommit = "0000000000000000000000000000000000000000"
 
 type (
+	// CreateBranch provides a SHA for creating a branch.
+	CreateBranch struct {
+		Ref string
```
bradrydzewski, line 18: just a question ... the branch will always have `refs/heads/` in the prefix.  Would it make sense to accept branch name, and then it would be the responsibility of the driver to determine whether or not to prefix with `refs/heads/`?  Similar to how find branch takes a branch name, as opposed to a full ref, and the underlying driver has to decide whether or not to prefix.

scm/driver/github/git.go
```diff
@@ -37,6 +37,15 @@ func (s *gitService) FindTag(ctx context.Context, repo, name string) (*scm.Refer
 	return convertRef(out), res, err
 }
 
+func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.CreateBranch) (*scm.Response, error) {
+	path := fmt.Sprintf("repos/%s/git/refs", repo)
+	in := &createBranch{
+		Ref: fmt.Sprintf("refs/heads/%s", params.Name),
```
bradrydzewski, line 43: we have a helper function that might be useful here:

```
scm.ExpandRef("refs/heads", params.Name)
```

tphoney: change made, thanks for the reminder. 👍

bradrydzewski (approved):

PR #108, opened 2021-05-15
Fixing Gitea commit API in case `ref/heads/` prefix is added to ref

Vici37:
If `FindCommit` gets used with a branch name instead of a sha, it may include the `ref/heads/` prefix, and the resulting API path of `.../commits/ref%2Fheads%2Fmy-branch` will return a 404. Trimming off that prefix so it's just `.../commits/my-branch` correctly returns the expected commit data.


eoinmcafee00:
Thank you @Vici37 . Merging this now.

PR #112, opened 2021-06-10
fix stash driver fetching repo permissions

### LINKED ISSUE #87: Drone and Bitbucket broken for write permission detection for drone build restart permission.
Since this commit:
https://github.com/drone/go-scm/commit/802e265b418f445ceec60827537bb5c7071cb018

Changing of 
	path := fmt.Sprintf("rest/api/1.0/repos?size=1000&permission=REPO_WRITE&project=%s&name=%s", namespace, name)
to 
	path := fmt.Sprintf("rest/api/1.0/repos?size=1000&permission=REPO_WRITE&projectname=%s&name=%s", namespace, name)

Has resulted in drone not allowing users with write permission to restart drone jobs or do drone deploys. As it is not detecting the user has write permission.

Admins still work using the webhook lookup but users with only write permission can not.

pedroparraortega:
tested with bitbucket server 7.6.0  check https://discourse.drone.io/t/bitbucket-credentials-not-properly-fetched-after-upgrade/9246/4

Fix #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 1.15.0
 ### Added
 - add delete file for github and gitlab, from [@tphoney](https://github.com/tphoney). See [#110](https://github.com/drone/go-scm/pull/110).
-- 
+
 ## 1.14.1
 ### Fixed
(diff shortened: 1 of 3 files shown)


tphoney:
Hey @pedroparraortega thanks for the fix, unfortunately it was a little more nuanced. The full fix is in #114 If you want to check it out. We are also doing a release today of the library. 

Thanks again ! 👍

PR #114, opened 2021-06-17
fix issue where write permission on bitbucket server wasn't working a…

eoinmcafee00:
…t repo level

### DIFF
--- a/scm/driver/stash/repo.go
+++ b/scm/driver/stash/repo.go
@@ -137,10 +137,10 @@ func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Pe
 		}, nil, nil
 	}
 	// HACK: test if the user has write access to the repository.
-	_, name := scm.Split(repo)
+	namespace, _ := scm.Split(repo)
 	repos, _, _ := s.listWrite(ctx, repo)
 	for _, repo := range repos {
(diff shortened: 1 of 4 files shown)


scm/driver/stash/repo.go
```diff
@@ -137,10 +137,10 @@ func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Pe
 		}, nil, nil
 	}
 	// HACK: test if the user has write access to the repository.
-	_, name := scm.Split(repo)
+	namespace, _ := scm.Split(repo)
 	repos, _, _ := s.listWrite(ctx, repo)
 	for _, repo := range repos {
-		if repo.Name == name {
+		if repo.Namespace == namespace {
```
tphoney, line 143: this code fix looks cool, one thing is that this was a bug where by the namespace and key were different, can we make sure that this is checked in the unit test https://github.com/drone/go-scm/blob/master/scm/driver/stash/repo_test.go#L142-L186

tphoney (changes requested):

scm/driver/stash/repo.go
eoinmcafee00: Done.

tphoney (approved):

PR #117, opened 2021-07-06
Support github pull request merge action

resec:


resec:
The pr test data json seems incorrect, when a pr is not merged `merged==false`, the `merged_at` field should be null.

Which causing the all the 3 cases are failing.

bradrydzewski:
> The pr test data json seems incorrect, when a pr is not merged merged==false, the merged_at field should be null.

I cannot speak to the implementation, but I can confirm the webhook payloads in the testdata folder are real payloads that were captured directly from github.

scm/driver/github/testdata/pr.json
```diff
@@ -371,7 +371,7 @@
         "site_admin": false
     },
     "merge_commit_sha": "e5bd3914e2e596debea16f433f57875b5b90bcd6",
-    "merged": false,
+    "merged": true,
```
bradrydzewski, line 374: the json payload in `pr.json` is for an open pull request;  therefore it should not be set to `merged: true`.  I would instead suggest adding a new json, real world json file and creating a new test case.

bradrydzewski (changes requested):

scm/driver/github/pr.go
```diff
@@ -154,7 +155,7 @@ func convertPullRequest(from *pr) *scm.PullRequest {
 		Link:   from.HTMLURL,
 		Diff:   from.DiffURL,
 		Closed: from.State != "open",
-		Merged: from.MergedAt.String != "",
```
bradrydzewski, line 157: can you provide some more details as to why this does not work?  I cannot recall, but there was a reason we did not use `merged` (perhaps compatibility with older versions of github enterprise, but I honestly cannot recall)

note that MergedAt handles null values and will interpret a null value as an empty string, so I would would expect this existing logic to work.

scm/driver/github/testdata/pr.json
resec: I tested on my own repo, the `merged` and `merge_at` are always marked or filled in the case of merged-pr. And both not marked or filled in the case of closed-pr and opening-pr.

scm/driver/github/pr.go
resec: The MergedAt is also workable, but I personally think the Merged field is more explicit.

resec: Because it is explicitly stated on the github doc.

scm/driver/github/testdata/pr.json
bradrydzewski: I think there may be a misunderstanding.  The `pr.json` file is for an open pull request (the `state` field is set to `open`) which means it cannot have merged set to true.  The `pr.json` file is a valid, real world json and this pull request makes the file invalid.

resec: Hmmm, it the `pr.json` is for a opening-pr, then its `merge_at` must be `null`. If it is not merged (opening), how can it have a merge time... So I think something in the `pr.json` must be wrong.

But having another test json (maybe `pr_merged.json`) make sense.

scm/driver/github/pr.go
bradrydzewski: are you certain this will not break older version of github enterprise?  we have to consider strict backward compatibility when we use this library.  If we were writing from scratch today we could use `merged`, but we have a number of github enterprise customers using this library and many of them are using older versions of github enterprise.

resec: ok, I will revert this change.

PR #120, opened 2021-09-09
[proposal/gitea driver] cosider all repo in private organization as private

a1012112796:
Hello, I found only when the repo is private or
set `DRONE_GIT_ALWAYS_AUTH`, drone will
use authenticate. but for a public repo
in private org, authenticate is also
necessary. so I think it's necessary
to cosider all repo in private organization
as private, thanks.


bradrydzewski:
:wave: thanks for the pull request.  A few things to get this merged.  First is that the unit tests are failing which means the implementation might be slightly flawed. Second is that we should handle the case where `src.Owner.Visibility` is both an empty value and a null value (since the latter will currently panic the code).

tphoney:
closing due to inactivity

PR #123, opened 2021-10-05
Update GitHub signature header to use sha256

nlecoy:
Hi !

First of all, thanks for the work done on Drone.

Just a quick PR to update the header used to validate webhooks from GitHub.
https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks#validating-payloads-from-github

GitHub still supports `X-Hub-Signature` for backward-compatibility but recommend using the `X-Hub-Signature-256`

### DIFF
--- a/scm/driver/github/webhook.go
+++ b/scm/driver/github/webhook.go
@@ -64,6 +64,9 @@ func (s *webhookService) Parse(req *http.Request, fn scm.SecretFunc) (scm.Webhoo
 	}
 
 	sig := req.Header.Get("X-Hub-Signature-256")
+	if sig == "" {
+		sig = req.Header.Get("X-Hub-Signature")
+	}
 	if !hmac.ValidatePrefix(data, []byte(key), sig) {
 		return hook, scm.ErrSignatureInvalid
 	}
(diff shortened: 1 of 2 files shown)


scm/driver/github/webhook.go
```diff
@@ -63,7 +63,7 @@ func (s *webhookService) Parse(req *http.Request, fn scm.SecretFunc) (scm.Webhoo
 		return hook, nil
 	}
 
-	sig := req.Header.Get("X-Hub-Signature")
+	sig := req.Header.Get("X-Hub-Signature-256")
```
bradrydzewski, line 66: We have some folks using this library with older versions of GitHub Enterprise.  In the interest of backward compatibility and ensuring we don't cause any regressions (for our Drone users) I am thinking we should fallback to the original header.

```Go
sig := req.Header.Get("X-Hub-Signature-256")
if sig == "" {
  sig = req.Header.Get("X-Hub-Signature")
}
```

bradrydzewski (changes requested):

nlecoy:
Hi there, 

Sorry to comme back to you just now (didn't see first answer). I added your suggestion @bradrydzewski and added some test.

bradrydzewski (approved):

tphoney:
Hey @nlecoy thanks for the PR then the additional work. !

PR #124, opened 2021-10-13
Feat: implemented gitee provider

kit101:
implemented gitee provider and unit test


scm/driver/gitee/git.go
```diff
@@ -0,0 +1,301 @@
+// Copyright 2017 Drone.IO Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gitee
+
+import (
+	"context"
+	"fmt"
+	"github.com/drone/go-scm/scm"
```
eoinmcafee00, line 10: Can you organise your imports please

With standard at the top,  drone imports in the middle and 3rd party at the bottom

kit101: resolved it.

PR #127, opened 2021-11-19
swap repo and target in bitbucket CompareChanges

jimsheldon:
bitbucket's API doc states the diffstat spec:

"Note: This is the opposite of the order used in git diff."

see https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/diffstat/%7Bspec%7D

I was unable to use this function without this change

### DIFF
--- a/scm/driver/bitbucket/git.go
+++ b/scm/driver/bitbucket/git.go
@@ -81,7 +81,7 @@ func (s *gitService) ListChanges(ctx context.Context, repo, ref string, opts scm
 }
 
 func (s *gitService) CompareChanges(ctx context.Context, repo, source, target string, opts scm.ListOptions) ([]*scm.Change, *scm.Response, error) {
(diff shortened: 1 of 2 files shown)


jimsheldon:
This can also be verified with curl, this
```
curl https://api.bitbucket.org/2.0/repositories/atlassian/atlaskit/diffstat/dec26e0fe887167743c2b7e36531dedfeb6cd478..425863f9dbe56d70c8dcdbf2e4e0805e85591fcc
```
returns
```
{"pagelen":500,"values":[],"page":1,"size":0}
```
while this
```
curl https://api.bitbucket.org/2.0/repositories/atlassian/atlaskit/diffstat/425863f9dbe56d70c8dcdbf2e4e0805e85591fcc..dec26e0fe887167743c2b7e36531dedfeb6cd478
```
returns the expected result

PR #131, opened 2021-11-28
fixbug: gitee webhook parse

kit101:
set pull_request action: `scm.ActionSync`


d1wilko:
Merged - thanks @kit101 :)

PR #133, opened 2021-12-08
Add workaround for bitbucket branch name with slash

jamie-harness:
Issue: Bitbucket API does not work if the branch name has a "/" in there. Bitbucket side has some open bugs. But it haven't been resolved over 7 years: 

https://jira.atlassian.com/browse/BCLOUD-9969

https://jira.atlassian.com/browse/BCLOUD-20223

Solution: call branch with query parameter and get branch instead of calling directly in case where there's a / in ref(which can be branch name) or branchname

### DIFF
--- a/scm/driver/bitbucket/git.go
+++ b/scm/driver/bitbucket/git.go
@@ -7,6 +7,7 @@ package bitbucket
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/drone/go-scm/scm"
@@ -28,6 +29,12 @@ func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.
 }
 
 func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Reference, *scm.Response, error) {
+	if strings.Contains(name, "/") {
+		path := fmt.Sprintf("/2.0/repositories/%s/refs/branches?q=name=\"%s\"", repo, name)
+		out := new(branches)
+		res, err := s.client.do(ctx, "GET", path, nil, out)
+		return convertBranch(out.Values[0]), res, err
+	}
 	path := fmt.Sprintf("2.0/repositories/%s/refs/branches/%s", repo, name)
 	out := new(branch)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
@@ -35,6 +42,12 @@ func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Re
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	if strings.Contains(ref, "/") {
+		path := fmt.Sprintf("2.0/repositories/%s/refs/branches?q=name=\"%s\"", repo, ref)
+		out := new(branches)
+		res, err := s.client.do(ctx, "GET", path, nil, out)
+		return convertCommit(out.Values[0].Target), res, err
+	}
 	path := fmt.Sprintf("2.0/repositories/%s/commit/%s", repo, ref)
 	out := new(commit)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
@@ -91,9 +104,7 @@ func (s *gitService) CompareChanges(ctx context.Context, repo, source, target st
 type branch struct {
 	Type   string `json:"type"`
 	Name   string `json:"name"`
-	Target struct {
-		Hash string `json:"hash"`
-	} `json:"target"`
+	Target *commit
 }
 
 type createBranch struct {
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/drone/go-scm
 
 require (
-	github.com/google/go-cmp v0.2.0
+	github.com/google/go-cmp v0.5.6
 	github.com/h2non/gock v1.0.9
 	github.com/nbio/st v0.0.0-20140626010706-e9e8d9816f32 // indirect
 )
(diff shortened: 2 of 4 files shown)


scm/driver/bitbucket/git.go
```diff
@@ -28,13 +29,25 @@ func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.
 }
 
 func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Reference, *scm.Response, error) {
+	if strings.Contains(name, "/") {
+		path := fmt.Sprintf("/2.0/repositories/%s/refs/branches?q=name=\"%s\"", repo, name)
```
bradrydzewski, line 33: you can use `%q` to quote

scm/driver/bitbucket/git.go
```diff
@@ -28,13 +29,25 @@ func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.
...
 	res, err := s.client.do(ctx, "GET", path, nil, out)
 	return convertBranch(out), res, err
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	if strings.Contains(ref, "/") {
+		path := fmt.Sprintf("2.0/repositories/%s/refs/branches?q=name=\"%s\"", repo, ref)
+		out := new(branches)
+		res, err := s.client.do(ctx, "GET", path, nil, out)
+		return convertCommit(out.Values[0].Target), res, err
```
bradrydzewski, line 49: if there is an error, `out.Values[0]` would result in an index out of bounds exception and a panic.  Also are we sure the first branch in the list is going to be the correct branch?

scm/driver/bitbucket/git.go
```diff
@@ -91,9 +104,7 @@ func (s *gitService) CompareChanges(ctx context.Context, repo, source, target st
 type branch struct {
 	Type   string `json:"type"`
 	Name   string `json:"name"`
-	Target struct {
-		Hash string `json:"hash"`
-	} `json:"target"`
+	Target *commit
```
bradrydzewski, line 107: do we need to change this?  in general, we try to avoid pointers in this library.  If we are going to introduce a pointer, we need to find everywhere the pointer is referenced and check for a null value. So if we don't need to change, I'd recommend leaving inline.

scm/driver/bitbucket/git.go
```diff
@@ -28,13 +29,25 @@ func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.
 }
 
 func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Reference, *scm.Response, error) {
+	if strings.Contains(name, "/") {
+		path := fmt.Sprintf("/2.0/repositories/%s/refs/branches?q=name=\"%s\"", repo, name)
+		out := new(branches)
+		res, err := s.client.do(ctx, "GET", path, nil, out)
+		return convertBranch(out.Values[0]), res, err
```
bradrydzewski, line 36: `out.Values[0]` has potential for panic on index out of bounds

scm/driver/bitbucket/git.go
jamie-harness: This is needed for convertCommit where it requires it the be the commit. I can write another convertCommit if it is better?

scm/driver/bitbucket/git.go
jamie-harness: If the branch Name is wrong then this will be nulltr. I can wrap the error and do the name check?

jamie-harness: The branch list is a name exact match so there should be exactly one. I will add error handling if it is not 1

scm/driver/bitbucket/git.go
```diff
@@ -28,13 +29,25 @@ func (s *gitService) CreateBranch(ctx context.Context, repo string, params *scm.
...
+	}
 	path := fmt.Sprintf("2.0/repositories/%s/refs/branches/%s", repo, name)
 	out := new(branch)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
 	return convertBranch(out), res, err
 }
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+	if strings.Contains(ref, "/") {
```
bradrydzewski, line 45: `FindCommit` is meant to accept a commit sha as opposed to a branch, so I'm not sure we need to make this change. If one is trying to lookup the commit for a branch, they should be using `FindBranch`

bradrydzewski:
I discussed with Jamie offline propose that we make two separate API calls.  First we make an API call to find the branch sha and then we make a second call to find the commit details.  Like this:

```Go
branch, _, err := client.Git.FindBranch(ctx, name)
if err != nil {
  panic(err)
}

commit, _, err := client.Git.FindCommit(ctx, branch.Sha)
if err != nil {
  panic(err)
}
```

It is technically possible to invoke `FindCommit` with a branch for GitHub and GitLab, however, this behavior is vendor-specific and does not work with other providers. Historically we try to abstract this library to the lowest common denominator, which is why we would recommend making two separate calls for all providers (even if it can be done in one single call).

If we decide we want to accept branch in `FindCommit` for all providers we could consider adding the below logic to the other providers, as needed. This probably merits a small discussion since it is more of a philosophical design question than a technical question, and because we would also need to account for tags, which require separate API calls and cannot be easily differentiated from branch names.

```diff
func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
+   if checkSha(ref) == false {
+       branch, resp, err := s.FindBranch(ctx, ref)
+       if err != nil {
+           return nil, resp, err
+       }
+       ref = branch.sha
+   }

```

I am going to recommend the workaround discussed above in the short term, and that we move this to a separate issue to decide whether or recommend two separate function calls (vendor neutral behavior) or support a single function call and abstract away the need for two API calls for non-GitHub and non-GitLab use cases.

PR #139, opened 2022-01-01
Add the support for delete of the bitbucket file

DeepakPatankar:


abhinav-harness (approved):

eoinmcafee00 (approved):

bradrydzewski:
I just have a few minor questions / comments.

I was wondering if we need to allocate a new string reader or if we can use `io.WriteString` or the built-in `WriteField` function? I noticed TP had used WriteField in the create and update functions he wrote.

```diff
-_, err = io.Copy(fw, strings.NewReader(content.File))
+_, err = fw.WriteField("message", content.Message)
```

Do we need to extract the bytes from the buffer and create a new reader or can we just pass the buffer directly to the request body?

```diff
-req.Body = bytes.NewReader(body.Bytes())
+req.Body = body
```

I noticed that the create and update endpoint checks to see if branch, sha and author are empty. Do we need the same check for the delete endpoint?  cc @tphoney what did you come across in your research when you implemented create and update?  See <https://github.com/DeepakPatankar/go-scm/blob/c81e22b5b929098137ce350ab31e1592becc1ea0/scm/driver/bitbucket/bitbucket.go#L85>

It also looks like we could improve the unit test which currently does not verify the request body in the gock check.  It would be nice to have a unit test to ensure future changes to this code don't cause a regression.  It also looks like this is an area of improvement for create and update contents endpoint (out of scope of this pull request, but would be nice to improve in a separate pr).

PR #143, opened 2022-01-12
Added support for parsing prevFilePath field from github compare commits API response

mohitg0795:
In case of RENAME ops, it returns prevFilePath of the file that has been renamed. So, added support to parse that field as part of go-scm API response.


tphoney (changes requested):
could you look at fixing the CI failure, 
and look to see if the unit tests can be extended.
thanks for the change

eoinmcafee00 (approved):

PR #145, opened 2022-01-15
Support to add the webhook with all scope on GitHub

LinuxSuRen:


bradrydzewski:
If you would like to add all events to a webhook for GitHub, this is something you can achieve today by providing all events to the `NativeEvents` field, for example:

```
input := &HookInput {
  NativeEvents: []string{"check_run", "check_suite", "code_scanning_alert", ... },
}
```

LinuxSuRen:
hi @bradrydzewski , thanks for your response. I know we can do it as you suggested. But it always needs to set a lot of event names. I'm wondering if we can make it be easier.

bradrydzewski:
I think one issue that would prevent this pull request from being merged is that it is only for GitHub.  The main purpose of this library is to abstract features across multiple vendors, and we have historically rejected pull requests that are not multi-vendor.  I would not envision us making an exception in this particular case.

If we added support for multiple vendors (github, gitlab, bitbucket server, bitbucket cloud, gitea) we would need to gauge demand for this feature before accepting the maintenance burden of agreeing to keep an up-to-date list of all event types across all providers.

LinuxSuRen:
Thanks for your time. I will try to add more vendors support.

> Is this something other folks need or would you be the only user?

I'm thinking about putting this feature into [this project](https://github.com/kubesphere/ks-devops/). It's not a personal project.

LinuxSuRen:
hi @bradrydzewski, please feel free to help to review it. Thanks for your time.

bradrydzewski:
for reasons described in my previous comment, we decided that this is out of scope for what we are willing to maintain.  Instead the caller should use the native events field to pass in all events.

> I suspect there will be a fair amount of burden for the maintainers to keep an up-to-date list of all event types, across all providers. I would not want to merge something unless we can commit to future maintenance.

PR #148, opened 2022-02-07
Extending bitbucket listCommits API to fetch commits for a given file

mohitg0795:


tphoney:
hey @mohitg0795 you are making a fundamental change to one of the base types for the library. In this case we would need to implement this change across all git providers eg github, gitea... 
Secondly with this additional parameter I would expect corresponding unit tests for this parameter, in each git provider.

tphoney:
For github https://docs.github.com/en/rest/reference/commits#list-commits--parameters
For gitlab https://docs.gitlab.com/ee/api/commits.html#list-repository-commits
For stash https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8283059808 - not possible. opts not implemented
For gitea https://gitea.com/api/swagger#/repository/repoGetAllCommits - not possible. opts not implemented

PR #154, opened 2022-02-21
Bitbucket list files fix

mohitg0795:
Currently, we only return first page of list files for bitbucket. It doesn't respect the incoming page number argument.

Bitbucket works on URL based pagination. Thus, added support if incoming page request has url, then it should be used, otherwise first page response will be returned.

### DIFF
--- a/scm/driver/bitbucket/content.go
+++ b/scm/driver/bitbucket/content.go
@@ -83,6 +83,10 @@ func (s *contentService) Delete(ctx context.Context, repo, path string, params *
 
 func (s *contentService) List(ctx context.Context, repo, path, ref string, opts scm.ListOptions) ([]*scm.ContentInfo, *scm.Response, error) {
 	endpoint := fmt.Sprintf("/2.0/repositories/%s/src/%s/%s?%s", repo, ref, path, encodeListOptions(opts))
+	if opts.URL != "" {
+		endpoint = opts.URL
+	}
+
 	out := new(contents)
 	res, err := s.client.do(ctx, "GET", endpoint, nil, out)
 	copyPagination(out.pagination, res)
--- a/scm/driver/bitbucket/content_test.go
+++ b/scm/driver/bitbucket/content_test.go
@@ -240,3 +240,29 @@ func TestContentList(t *testing.T) {
 		t.Log(diff)
 	}
 }
+
+func TestContentListWithUrlInput(t *testing.T) {
+	defer gock.Off()
+
+	mockNextPageUri := "https://api.bitbucket.org/2.0/repositories/atlassian/atlaskit/src/master/packages/activity?pageLen=3&page=RPfL"
+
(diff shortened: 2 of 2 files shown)


scm/driver/bitbucket/content.go
```diff
@@ -83,6 +83,10 @@ func (s *contentService) Delete(ctx context.Context, repo, path string, params *
 
 func (s *contentService) List(ctx context.Context, repo, path, ref string, opts scm.ListOptions) ([]*scm.ContentInfo, *scm.Response, error) {
 	endpoint := fmt.Sprintf("/2.0/repositories/%s/src/%s/%s?%s", repo, ref, path, encodeListOptions(opts))
+	if opts.URL != "" {
```
tphoney, line 86: This is incorrect, if you look at the encodeListOptions function it allows the user to pass through a page number https://github.com/mohitg0795/go-scm/blob/21de0f1a1424e92ac3a2c34aa26519a9f1f464d2/scm/driver/bitbucket/util.go#L27

tphoney (changes requested):

scm/driver/bitbucket/content.go
mohitg0795: It doesn't get respected for bitbucket. Bitbucket doesn't support page number based pagination. It only supports url based pagination. Whatever page number you may pass, it will always provide you first page of output.

Thus, not tinkering with current logic, I just added support for url based pagination. If someone provides page url, then it overrides everything and it is used to parse the results.

tphoney: can you provide an example of the url you would pass. 
i am extremely hesitant to merge this kind of change. as the URL parameter ignores all other parameters. at this point it would be better to use a simple http request. 
this makes using bitbucket totally different to other git providers, meaning end users will miss this nuance.
We are having a team meeting later and will discuss this PR then

tphoney: https://developer.atlassian.com/cloud/bitbucket/rest/api-group-source/#directory-listings

mohitg0795: This is a sample URL.
`mockNextPageUri := "https://api.bitbucket.org/2.0/repositories/atlassian/atlaskit/src/master/packages/activity?pageLen=3&page=RPfL"
`

Also, I've tested it by integrating with bitbucket realtime and got this URL during that process.

Further, I'm not tampering any behaviour. I'm just enhancing current support and infact fixing current behaviour. There's no page num based support for bitbucket. It results into HTTP 500 if you pass any page number except 0. You can further read on bitbucket documentation around this API. It only supports URL based pagination. It returns next page complete URL when we try to fetch 0th page and consequently it just works recursively in same way for other pages.

Let me know if we need to come on a call to finalize on this change. Appreciate your help. Thanks.

tphoney:
After the team review, this is the best solution available. Due to bitbucket.

PR #156, opened 2022-02-25
(DRON-242) initial implementation of azure devops support

tphoney:


scm/client.go
```diff
@@ -86,6 +86,9 @@ type (
 		// Base URL for API requests.
 		BaseURL *url.URL
 
+		// project ID only used for azure devops.
+		AzureRepoID string
```
bradrydzewski, line 90: this can be moved to `wrapper` so that we don't need to leak any driver specific details into the main scm package

scm/driver/azure/azure.go
```diff
@@ -0,0 +1,109 @@
...
+func NewDefault(azureOwner, azureProject string) *scm.Client {
+	client, _ := New("https://dev.azure.com", azureOwner, azureProject)
+	return client
+}
+
+// wrapper wraps the Client to provide high level helper functions for making http requests and unmarshaling the response.
+type wrapper struct {
+	*scm.Client
+	AzureOwner   string
+	AzureProject string
```
bradrydzewski, line 62: could these be changed to private variables?

scm/driver/azure/azure.go
```diff
@@ -0,0 +1,109 @@
...
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/drone/go-scm/scm"
+)
+
+// New returns a new azure API client.
+func New(uri, azureOwner, azureProject string) (*scm.Client, error) {
```
bradrydzewski, line 20: we could probably rename to `owner` and `project`, where `azure` is inferred from the package name.

tphoney: done

scm/driver/azure/azure.go
tphoney: done

scm/driver/azure/integration/content_test.go
```diff
@@ -0,0 +1,150 @@
...
+	repoID       = "fde2d21f-13b9-4864-a995-83329045289a"
+)
+
+func TestContentsFind(t *testing.T) {
+	if token == "" {
+		t.Skip("Skipping, Integration test")
+	}
+	client = azure.NewDefault(organization, project)
+	client.Client = &http.Client{
+		Transport: &transport.Custom{
```
bradrydzewski, line 30: Just a minor note that you can use `transport.BearerToken{ Token: token }`

PR #168, opened 2022-04-20
(fix) handle nil repos in github responses

tphoney:


bradrydzewski:
another option could be to move the nil check into the convert list function, and prevent nil repositories from being appended to the list all together

```diff
// helper function to convert from the gogs repository list to
// the common repository structure.
func convertRepositoryList(from []*repository) []*scm.Repository {
	to := []*scm.Repository{}
	for _, v := range from {
+		if v != nil {
			to = append(to, convertRepository(v))
+		}
	}
	return to
}
```

scm/driver/github/repo.go
```diff
@@ -187,14 +187,20 @@ func (s *RepositoryService) DeleteHook(ctx context.Context, repo, id string) (*s
 func convertRepositoryList(from []*repository) []*scm.Repository {
 	to := []*scm.Repository{}
 	for _, v := range from {
-		to = append(to, convertRepository(v))
+		repo := convertRepository(v)
```
bradrydzewski, line 190: we can condense

```
if repo := convertRepository(v); repo != nil {
  to = append(to, repo)
}
```

eoinmcafee00 (approved):

PR #173, opened 2022-05-02
[PL-24911]: Make project name as optional param in Azure Repo APIs

DeepakPatankar:
For the list Repo api of the Azure repo. the project name is not a required parameter.In Harness, we support an Account type connectors which won’t contain the project name. When we are doing the test connection of this connector, we call the scm API to list all the repositories of an organization.  Currently the scm code doesn’t allow the project name as null, which I fixed in this ticket

### DIFF
--- a/scm/driver/azure/azure.go
+++ b/scm/driver/azure/azure.go
@@ -25,7 +25,7 @@ func New(uri, owner, project string) (*scm.Client, error) {
 	if !strings.HasSuffix(base.Path, "/") {
 		base.Path = base.Path + "/"
 	}
-	if owner == "" || project == "" {
+	if owner == "" {
 		return nil, fmt.Errorf("azure owner and azure project are required")
 	}
 	client := &wrapper{
--- a/scm/driver/azure/azure.go
+++ b/scm/driver/azure/azure.go
@@ -26,7 +26,7 @@ func New(uri, owner, project string) (*scm.Client, error) {
 		base.Path = base.Path + "/"
 	}
 	if owner == "" {
-		return nil, fmt.Errorf("azure owner and azure project are required")
+		return nil, fmt.Errorf("azure owner is required")
 	}
 	client := &wrapper{
 		new(scm.Client),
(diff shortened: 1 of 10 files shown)


scm/driver/azure/azure.go
```diff
@@ -25,7 +25,7 @@ func New(uri, owner, project string) (*scm.Client, error) {
 	if !strings.HasSuffix(base.Path, "/") {
 		base.Path = base.Path + "/"
 	}
-	if owner == "" || project == "" {
+	if owner == "" {
 		return nil, fmt.Errorf("azure owner and azure project are required")
```
mohitg0795, line 29: logline should be fixed accordingly.

tphoney (approved):

tphoney:
@DeepakPatankar looks good with the change, but you will need to fix the unit tests first.

scm/driver/azure/repo.go
```diff
@@ -40,7 +40,12 @@ func (s *RepositoryService) FindPerms(ctx context.Context, repo string) (*scm.Pe
 // List returns the user repository list.
 func (s *RepositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
 	// https://docs.microsoft.com/en-us/rest/api/azure/devops/git/repositories/list?view=azure-devops-rest-6.0
-	endpoint := fmt.Sprintf("%s/%s/_apis/git/repositories?api-version=6.0", s.client.owner, s.client.project)
+	var endpoint string
+	if s.client.project != "" {
```
allofthesepeople, line 44: nit: prefer positive to negative conditionals as they're easier/quicker to understand.

allofthesepeople (approved):
minor nit on conditional

scm/driver/azure/repo.go
DeepakPatankar: done

scm/driver/azure/azure.go
DeepakPatankar: done

tphoney (commented):
If a user creates a scm client and does not set the project, it will fail spectacularly for all calls except for List repositories. It does not feel like there is any protection against this. with the change to 
```
if owner == "" || project == "" {
		return nil, fmt.Errorf("azure owner and azure project are required")
	}
```

PR #175, opened 2022-05-06
fix the duplication of repos in traverse call

aman-harness:


scm/traverse/repos.go
```diff
@@ -20,10 +20,8 @@ func Repos(ctx context.Context, client *scm.Client) ([]*scm.Repository, error) {
 		if err != nil {
 			return nil, err
 		}
-		for _, src := range result {
-			if src != nil {
-				list = append(list, result...)
-			}
+		if result != nil {
```
shubham149, line 23: result won't be nil ever

scm/traverse/repos.go
```diff
@@ -20,10 +20,8 @@ func Repos(ctx context.Context, client *scm.Client) ([]*scm.Repository, error) {
 		if err != nil {
 			return nil, err
 		}
-		for _, src := range result {
-			if src != nil {
-				list = append(list, result...)
```
shubham149, line 25: Change `result` to `src` in this line. Intended code was to add the repo if it is not nil

bradrydzewski:
good catch @shubham149 ... fixed by https://github.com/drone/go-scm/commit/4c30731be442baccbbfd47bfd974f711d4b9788c

PR #181, opened 2022-05-11
[PL-24913]: Handle the error raised while creating a multipart input

DeepakPatankar:


scm/multipart.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (mw *MultipartWriter) Close() {
+	mw.Writer.Close()
+}
+
+func (mw *MultipartWriter) FormDataContentType() string {
+	return mw.Writer.FormDataContentType()
+}
+
+type MultipartWriter struct {
```
allofthesepeople, line 23: nit: typically the struct definition comes before the methods.

scm/driver/stash/stash.go
```diff
@@ -79,26 +79,29 @@ func (c *wrapper) do(ctx context.Context, method, path string, in, out interface
 		case *contentCreateUpdate:
 			// add the content to the multipart
 			var b bytes.Buffer
-			w := multipart.NewWriter(&b)
+			mw := &scm.MultipartWriter{Writer: multipart.NewWriter(&b)}
 			// add the other fields
-			if content.Message != "" {
-				_ = w.WriteField("content", string(content.Content))
+			if string(content.Content) != "" {
```
allofthesepeople, line 84: the empty string checks are handled within the `mw.Write` method, so could probably lose these additional checks. 

Though see comment below on whether the general purpose struct should be checking for empty values

scm/multipart.go
```diff
@@ -0,0 +1,26 @@
+package scm
+
+import "mime/multipart"
+
+func (mw *MultipartWriter) Write(f, v string) {
+	if mw.Error == nil {
+		return
+	}
+	if v == "" {
```
allofthesepeople, line 9: Should possibly leave a note in the method comment about empty strings not being written as this differs from the mime/multipart behaviour it's wrapping.

allofthesepeople (approved):
couple of nits but nice simplification.

scm/driver/stash/stash.go
DeepakPatankar: done

scm/multipart.go
DeepakPatankar: done

scm/multipart.go
DeepakPatankar: done

scm/multipart.go
```diff
@@ -0,0 +1,26 @@
+package scm
```
tphoney, line 1: So this functionality is only used by stash/content. I would move this helper method in under stash 
or 
into scm/driver/stash/stash.go

tphoney (changes requested):

scm/multipart.go
tphoney:
```
➜  go-scm git:(no_project) grep -irsn Multipart .
./scm/driver/bitbucket/bitbucket.go:13: "mime/multipart"
./scm/driver/bitbucket/bitbucket.go:74:                 // add the content to the multipart
./scm/driver/bitbucket/bitbucket.go:77:                 w := multipart.NewWriter(&b)
./scm/driver/bitbucket/bitbucket.go:95:                 // write the multipart response to the body
./scm/driver/bitbucket/bitbucket.go:97:                 // write the content type that contains the length of the multipart
./scm/driver/bitbucket/bitbucket.go:103:                        writer := multipart.NewWriter(body)
```
nothing else is using it yet. I would definitely make it only for stash for now. Until another git provider uses it

tphoney (approved):

PR #194, opened 2022-06-14
Support parsing Gitlab Note Hook event

rutvijmehta-harness:


rutvijmehta-harness:
@shubham149 @tphoney can I please get a review for this change?

scm/driver/gitlab/webhook.go
```diff
@@ -202,6 +215,65 @@ func converBranchHook(src *pushHook) *scm.BranchHook {
...
+			Body:    src.MergeRequest.Description,
+			Ref:     fmt.Sprintf("refs/merge-requests/%d/head", src.MergeRequest.Iid),
+			Source:  src.MergeRequest.SourceBranch,
+			Target:  src.MergeRequest.TargetBranch,
+			Link:    src.Project.WebURL,
+			Closed:  src.MergeRequest.State != "opened",
+			Merged:  src.MergeRequest.State == "merged",
+			Author:  *convertUser(&src.User),
+			Created: parseTimeString(src.MergeRequest.CreatedAt),
+			Updated: parseTimeString(src.MergeRequest.UpdatedAt),
```
jamie-harness, line 238: Is there a way to get label and commits here? or it probably have to be a separate call

jamie-harness:
+ @bradrydzewski

scm/driver/gitlab/webhook.go
rutvijmehta-harness: Added labels and last commit Sha in the PR struct.

scm/driver/gitlab/webhook.go
```diff
@@ -380,93 +471,76 @@ type (
...
+			CommitID         string      `json:"commit_id"`
+			NoteableID       int         `json:"noteable_id"`
+			StDiff           interface{} `json:"st_diff"`
+			System           bool        `json:"system"`
+			ResolvedAt       interface{} `json:"resolved_at"`
+			ResolvedByID     interface{} `json:"resolved_by_id"`
+			ResolvedByPush   interface{} `json:"resolved_by_push"`
+			DiscussionID     string      `json:"discussion_id"`
+			URL              string      `json:"url"`
+			Position         interface{} `json:"position"`
```
shubham149, line 493: Why are we changing Position to an interface? Same for others as well

rutvijmehta-harness: There is a mismatch between the datatypes used for Position in this file previously and the gitlab go package https://github.com/go-playground/webhooks/blob/v5.17.0/gitlab/payload.go#L322. 

1. It doesn't contain extra fields which are present in Position https://github.com/go-playground/webhooks/blob/59fb74817c3d5c702553964f4d2faff80f23d350/gitlab/payload.go#L385 which can lead to unmarshal failing
2. Position/OriginalPosition/ChangedPosition have the same type in the go package but different here

Ideally we should keep the Position type same but kept interface here to be on the safer side. I could not find documentation on what the possible values could be each field.

shubham149 (approved):

PR #200, opened 2022-06-22
added issue comment hook support for Azure

raghavharness:


tphoney (changes requested):
Are there any other types of events for the comment webhook eg for github there are a number of different actions https://github.com/drone/go-scm/blob/37ec420b65c2c4cd544ad49817986c3ccdb41644/scm/driver/github/webhook.go#L132-L142

raghavharness:
@tphoney Added

scm/const.go
```diff
@@ -67,6 +67,8 @@ func (a Action) String() (s string) {
 		return "synchronized"
 	case ActionMerge:
 		return "merged"
+	case ActionEdit:
```
tphoney, line 70: is this not really as update and edit are synonyms
```
case ActionUpdate:
		return "updated"
```

tphoney (changes requested):

scm/const.go
raghavharness: @tphoney We can't keep this as Update since in triggers we have already defined an edit action on UI. We just checked since ActionEdit was not being passed Github IssueComment trigger with Edit action was not triggering so we need to pass this value. You can check below snippet.

https://github.com/drone/go-scm/blob/bff969507a83c53a7928825cc7c1c57be3c01b87/scm/driver/github/webhook.go#L135

tphoney, line 70: @bradrydzewski how do you feel about adding this new action, it does seem to be referenced in some of the drivers.

bradrydzewski: We should not need to add a new action here.  If this refers to updating pull request code we should use the `synchronized` action (which is the abstracted term we use for all providers) or if this refers to updating metadata (tltle, description, status, etc) we should use the `update` action.

See some of the existing implementations for prior art:
https://github.com/drone/go-scm/blob/master/scm/driver/github/webhook.go#L159:L160

bradrydzewski (changes requested):

scm/const.go
bradrydzewski: The purpose of this library is to provide an abstraction layer across all scm providers, so that we can use the same data structures and terminology for all providers.  This library is not intended to align with the Harness user interface, which may choose to expose scm-specific data.  This library should only be used where abstract is needed;   if you want the original data or do not need abstractions, you would not use this library.  Although perhaps I am misunderstanding.

tphoney (approved):

PR #204, opened 2022-07-07
Update sender login id for azure repo

raghavharness:


scm/driver/azure/webhook.go
```diff
@@ -335,7 +335,7 @@ func convertIssueCommentHook(src *issueCommentPullRequestHook) *scm.IssueComment
 		},
 		Sender: scm.User{
 			Email:   src.Resource.Comment.Author.UniqueName,
-			Login:   src.Resource.Comment.Author.ID,
+			Login:   src.Resource.Comment.Author.DisplayName,
```
shubham149, line 338: this is incorrect
Login has to set to ID. Login is a unique id while name can be duplicates. Name's value is set to Display name.

shubham149 (changes requested):

scm/driver/azure/webhook.go
raghavharness: I see, just got to know we need some unique identifier for licensing. The ID here is a UUID which we don't want show on UI. Azure provides another unique name parameter which we can use instead. I'll discuss this further.

PR #209, opened 2022-07-24
Fix Gogs find commit

wandersonwhcr:
This PR fixes `findCommit` from Gogs when reference is not a SHA1.

### DIFF
--- a/scm/driver/gogs/git.go
+++ b/scm/driver/gogs/git.go
@@ -29,8 +29,10 @@ func (s *gitService) FindBranch(ctx context.Context, repo, name string) (*scm.Re
 
 func (s *gitService) FindCommit(ctx context.Context, repo, ref string) (*scm.Commit, *scm.Response, error) {
 	if scm.IsHash(ref) == false {
-		if branch, _, err := s.FindBranch(ctx, repo, scm.TrimRef(ref)); err == nil {
-			ref = branch.Sha // replace ref with sha
+		path := fmt.Sprintf("api/v1/repos/%s/commits/%s", repo, ref)
+		res, err := s.client.do(ctx, "GET", path, nil, ref)
+		if err != nil {
+			return nil, res, err
 		}
 	}
 	path := fmt.Sprintf("api/v1/repos/%s/commits/%s", repo, ref)
--- a/scm/driver/gogs/git.go
+++ b/scm/driver/gogs/git.go
@@ -7,6 +7,7 @@ package gogs
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/drone/go-scm/scm"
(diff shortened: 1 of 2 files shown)


bradrydzewski:
👋  thanks for the pull request.  We implemented a similar solution for Bitbucket.  Can you please match the implementation / coding style here: <https://github.com/drone/go-scm/blob/master/scm/driver/bitbucket/git.go#L38:L46>

wandersonwhcr:
Wow! TY!

I am not a Golang programmer :laughing:

bradrydzewski:
fixed by https://github.com/drone/go-scm/commit/07b069c7f79af40bf6cdb2bd91022591477be0a6

wandersonwhcr:
@bradrydzewski I was trying to resolve and learn Go.

BTW, TY.

bradrydzewski:
> bradrydzewski I was trying to resolve and learn Go.

totally understand and appreciate ... it was blowing up my notifications so I decided to resolve.  We are happy to accept contributions, but perhaps would politely ask that you run your unit tests locally to minimize pull request update notifications :)

wandersonwhcr:
FYI, your tests define how GitLab works and not Gogs.

When I manually run a job on Drone, this is the request:

```
"GET /api/v1/repos/root/teste/commits/refs/heads/master HTTP/1.1" 500 88 "-" "Go-http-client/1.1" "-"
```

This is why I didn't used your example.

Good luck.

bradrydzewski:
Sorry, not sure I understand the comment. 

I confirmed that these API calls to Gogs are valid, using my local installation:

```shell
$> curl http://localhost:10880/api/v1/repos/foo/bar/branches/master

{
  "name": "master",
  "commit": {
    "id": "f05255b318f2985fa108cd3260295b86b388fe8a"
    ....
}

$> curl http://localhost:10880/api/v1/repos/foo/bar/commits/f05255b318f2985fa108cd3260295b86b388fe8a

{
  "name": "master",
  "commit": {
    "id": "f05255b318f2985fa108cd3260295b86b388fe8a",
    ....
}
```

The unit tests match the above calls (using slightly different data).

```Go
	gock.New("https://try.gogs.io").
		Get("/api/v1/repos/gogits/gogs/branches/master").
		Reply(200).
		Type("application/json").
		File("testdata/branch.json")

	gock.New("https://try.gogs.io").
		Get("/api/v1/repos/gogits/gogs/commits/f05f642b892d59a0a9ef6a31f6c905a24b5db13a").
		Reply(200).
		Type("application/json").
		File("testdata/commits.json")
```

The change I made will get the branch details if and only if the reference (input parameter) is not a sha.  This unit test verifies that if the reference is `master`, we make an API call to get the branch details for master, extract the sha from the json response, and use the sha to make an API call and get the commit details.

> When I manually run a job on Drone, this is the request:

Not sure I follow.  Drone would be using the old version of the code, before my change.  If you look at the new code, it uses `scm.TrimRef(ref)` to remove `refs/heads` which is what Gogs seems to be expecting.

ShruthiHubb:
👍

PR #212, opened 2022-07-29
Pl 26239

bhavya181:


bhavya181:
I have created a interface ListRepoRequestParams which will contain all the extra parameters as per git provider, each git provider will implement this interface and have different fields in it, and then listV2 method has been defined which takes this interafce as one of the parameter.

scm/driver/azure/repo.go
```diff
@@ -55,6 +58,10 @@ func (s *RepositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*
 	return convertRepositoryList(out), res, err
 }
 
+func (s *RepositoryService) ListV2(ctx context.Context, AzureRequestParams ListRepoRequestParams, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
```
bradrydzewski, line 61: should follow Go / project naming conventions

bradrydzewski:
I have some reservations.  The purpose of this repository is to abstract common functionality across all git providers (github, bitbucket, gitlab, etc).  This adds github-only functionality to all providers.  I also noticed the naming conventions and coding style do not match Go conventions, nor do they match conventions used throughout this repository.

bhavya181:
that's correct, the naming convention and coding style may not match with Go style, as its just a pseudocode to propose our approach, and coming on adding github-only functionality, we just want to abstract class which is query param to the method, so that each provider can put the extra ields, if they will require any in future for this listing api.
will take team's help for same. Thanks

tphoney:
I agree with @bradrydzewski changing the interface for all git providers is a bad approach and it makes it very confusing for non-harness users. 

My suggestion before https://github.com/drone/go-scm/compare/master...tphoney-patch-1?quick_pull=1 is still a better approach do you concur @bradrydzewski  ?

bhavya181:
@bradrydzewski so the aim is that, for github app there is a different end point for listing repos, so we just want to pass a boolean to understand where is github app or not in case of github, now according to @tphoney we should just pass that github app endpoint in listOptions.url, but then scm svc (harness-core) should pass an endpoint which should not be the repsonsibility of scm svc, ideally scm svc should just say we want repo list for github app, then go-scm should call the right end point to get details.

bradrydzewski:
@TP in this case, not sure the `opts.URL` approach is applicable here.  These are really two completely different functions, and people invoking `RepositoryService.List` may not expect this function to use the `installation/repositories` endpoint.  In this case, we are trying to expose GitHub-only functionality, which has historically been out of scope for this library.

I see two options.

1. Push this logic to the Harness scm-service to avoid polluting this library with code that only Harness needs
2. Come up with a design strategy for how we want to support vendor specific endpoint

---

If we want option 2, this could be done by adding an endpoint _only_ to the `github.RepositoryService`.  There is no need to add this endpoint to the other drivers, or to register this method with the `scm.RepositoryService` interface.

```diff
+func (s *RepositoryService) ListInstallation(ctx context.Context, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	path := fmt.Sprintf("installation/repositories?%s", encodeListOptions(opts))
+	out := []*repository{}
+	res, err := s.client.do(ctx, "GET", path, nil, &out)
+	return convertRepositoryList(out), res, err
+}
```

The caller can invoke GitHub-specific endpoint by casting to the correct type:

```
client.Repos.(*github.RepositoryService).ListInstallation(ctx, opts)
```

bhavya181:
@bradrydzewski  this approach looks good, we can go ahead with this approach, what's say @tphoney?

Got your point regarding library, we should expose specific listing method only in github.repositoryService

bradrydzewski:
> however this will significantly increase complexity in the scm service

we need to add complexity somewhere (delegate, scm service, scm library).  There are two options for how this could be implemented.  First is that the scm-service exposes both List and ListInstances and the delegate chooses which to invoke. Second is the delegate passes in a boolean flag to the List endpoint, and there is an `if` statement to decides which function to invoke.

either way, this keeps the scm library clean, which is the primary goal.  We are dealing with a lot of tech debt throughout, but this library has remained clean, and I'd like to keep it that way.

> If we do go with the approach of having a github application only method, that is good in terms of not polluting the go-scm service

the reason I am willing to consider github-specific methods is because we will probably soon need to support github checks, which is a github-only feature.

bhavya181:
@bradrydzewski  @tphoney, I think we all agree on adding a new method in go-scm and passing a boolean field from delegate to scm service, then scm service will make call based on boolean, should I go ahead with the implementation?

PR #218, opened 2022-08-23
[PL-26239]: fix for list response

bhavya181:
response schema for listUserRepositries and listAppnstallations was different, and that's why it was throwing error, so changes the output.
Tested with writing integartion test with my credentials, now, its working fine.


adivishy1 (approved):

scm/driver/github/repo.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/drone/go-scm/scm"
 )
 
-type repository struct {
+type Repository struct {
```
mohitg0795, line 17: From what I understand, it makes the struct public. Why do we need to make it public if all references are from same package/file?
Sorry, I may be naive here, just trying to understand its impact

bhavya181: json unmarshal and marshal only works when name of struct start with capital letter

mohitg0795: Oh interesting. Thanks for the info.

mohitg0795 (approved):

scm/driver/github/repo.go
tphoney: image
https://github.com/drone/go-scm/pull/219
repositories should be lower case as @mohitg0795 said

tphoney (changes requested):

scm/driver/github/repo.go
```diff
@@ -53,6 +53,11 @@ type hook struct {
 	} `json:"config"`
 }
 
+type RepositoryListResponse struct {
```
tphoney, line 56: should start with a lowercase letter

tphoney (changes requested):

scm/driver/github/repo.go
bhavya181: but then, RepositoryListResponse will not be able to map to Repository struct as it does not unmarshal that struct

bhavya181: Can you suggest any other method, so that it will map correctly

tphoney: can you show me where it doesnt work, the unit tests passed and it unmarshalled successfully

if you look at the do method for all drivers, it marshalls / unmarshalls without all of the structs. The other structs all start with a lowercase letter.

scm/driver/github/repo.go
bradrydzewski: agree this should be lowercase.   Also we do not suffix with `Response` so it should be `repositoryList` instead

bradrydzewski:
The `repository` struct should never be exposed as public, it should be kept private and converted to `scm.Repository`.  The purpose of this library is to convert vendor-specific structures to common abstractions.  The `scm.Repository` is the common abstract that must be used.

If the response structure is different, you should create a new `applicationRepositoryList` structure instead.  Then you would convert the repositories in the `applicationRepositoryList` to a `[]scm.Repository` structure.

scm/driver/github/repo.go
bhavya181: ya, I have changed them now

scm/driver/github/repo.go
bhavya181: changed

tphoney (approved):

PR #225, opened 2022-09-14
Bitbucket sha fix for merged pr

raghavharness:
For bitbucket, since we don't have pull-request refs so we use source branch to clone the PR. But in cases where source branch is deleted when pr is merged, git clone gets failed. To prevent that we'll use the merge commit instead of source branch latest commit.


bradrydzewski:
@raghavharness please keep in mind that there are multiple consumers of this library, and that changes made for Harness have the potential to break other consumers, such as Drone.  For example, when Drone executes a pipeline for a pull request, it clones the base branch and then manually performs the merge.  Moving from commit sha to merge sha would break the clone step in Drone for Bitbucket.

Instead of changing the behavior of an existing field, I have added a new `Merge` field that you can use to capture the merge sha. https://github.com/drone/go-scm/blob/master/scm/pr.go#L27

bradrydzewski (changes requested):
capture merge sha in `PullRequest.Merge` field instead

raghavharness:
Sure, thanks @bradrydzewski for the inputs.

raghavharness:
@bradrydzewski @tphoney Have addressed your comment, can you please review once. Thanks.

PR #234, opened 2023-01-27
feat: [PIE-7927]: added header to avoid/bypass csrf check

mohitg0795:
We got a customer issue that they are getting CSRF check issues while using bitbucket on-prem connector. They suggested to add a HTTP header to requests to bypass this issue. We read more on it and found it safe to add such a header. Adding "x-atlassian-token" header helps to bypass csrf checks which are of no significance while using APIs.

For more details, refer: https://community.sonarsource.com/t/bitbucket-server-integration-csrf-xsrf-warning-in-logs/23604/4


tphoney (approved):

tphoney:
Great PR, thanks for the research and explanation.

PR #256, opened 2023-05-12
feat: [CDS-69341]: add find user email api for github in go-scm

shalini-agr:


scm/driver/github/user.go
```diff
@@ -31,8 +31,9 @@ func (s *userService) FindLogin(ctx context.Context, login string) (*scm.User, *
 }
 
 func (s *userService) FindEmail(ctx context.Context) (string, *scm.Response, error) {
-	user, res, err := s.Find(ctx)
-	return user.Email, res, err
```
bradrydzewski, line 35: you can re-use the `ListEmail` function here:

```
func (s *userService) FindEmail(ctx context.Context) (string, *scm.Response, error) {
	out, res, err := s.ListEmails(ctx, scm.ListOptions{})
	return returnPrimaryEmail(out), res, err
}
```

scm/driver/github/user_test.go
```diff
@@ -87,14 +87,14 @@ func TestUserEmailFind(t *testing.T) {
...
-		Get("/user").
+		Get("/user/emails").
 		Reply(200).
 		Type("application/json").
 		SetHeader("X-GitHub-Request-Id", "DD0E:6011:12F21A8:1926790:5A2064E2").
 		SetHeader("X-RateLimit-Limit", "60").
 		SetHeader("X-RateLimit-Remaining", "59").
 		SetHeader("X-RateLimit-Reset", "1512076018").
-		File("testdata/user.json")
+		File("testdata/userEmail.json")
```
bradrydzewski, line 97: files in the testdata folder use snake case for file names as opposed to camel case (e.g. `user_email.json` as opposed to `userEmail.json`) however, you can probably re-use the

bradrydzewski: although you may be able to re-use the existing `testdata/emails.json` file in that folder for your test

scm/driver/github/user.go
shalini-agr: done

shalini-agr: done

scm/driver/github/user_test.go
shalini-agr: done

mohitg0795 (approved):

PR #260, opened 2023-07-05
[feat]: [CDS-73030]: Support for text based branch filtration

adivishy1:
Have added code for supporting text based branch filtration for the following git providers:

1. Bitbucket cloud
2. Bitbucket server
3. Azure
4. Gitlab

Note: Github does not provide support for text based search.
Have tested out string based search all the above providers. For the other drivers that implement the GitService interface, have called the listBranches method which means that its not filtering based on any branch being passed.

### DIFF
--- a/scm/client.go
+++ b/scm/client.go
@@ -68,6 +68,13 @@ type (
 		Reset     int64
 	}
 
+	// BranchListOptions specifies optional branch search term and pagination
+	// parameters.
+	BranchListOptions struct {
+		SearchTerm      string
+		PageListOptions ListOptions
+	}
+
 	// ListOptions specifies optional pagination
 	// parameters.
 	ListOptions struct {
--- a/scm/driver/azure/git.go
+++ b/scm/driver/azure/git.go
@@ -63,6 +63,17 @@ func (s *gitService) ListBranches(ctx context.Context, repo string, _ scm.ListOp
 	return convertBranchList(out.Value), res, err
 }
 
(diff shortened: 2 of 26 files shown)


mohitg0795 (approved):

raghavharness (approved):

tphoney (approved):
i have a few questions. 
what do you expect the search term to look like ? do we need to encode that search term differently for each scm provider ?
why dont you just use the url operation that is part of the listBranches ?
why is it named v2 , this has no meaning

adivishy1:
1. what do you expect the search term to look like ? : it can be anything that user types, say if we have branches such as test, main, main-patch, main-patch1, main-patch-12 cds-12312, etc, and the user wants only branches where "main-patch" in present, then in such case the user can pass search term as "main-patch" or "patch" or "main-" the the git providers that support such functionality 
2. do we need to encode that search term differently for each scm provider ? yes, each of them take the filter params in a different query param name, eg azure takes it as `"filterContains="` ,bitbucket saas takes it as `"q=name~"` etc,
3. why dont you just use the url operation that is part of the listBranches?: there are cases where the url changes slightly, like in the case of azure, in existing impl of ListBranches, there is a query param `includeMyBranches=true`  but we cannot have `includeMyBranches` and `"filterContains="`  together in the same url, hence new listBranchesV2 with new query params in the url
4. why is it named v2 , this has no meaning: Purely for backward compatibility purpose, for implementing this functionality, we would need a new param, searchTerm to the existing listBranches method, if we directly add it as new param, that would break the previous usages of the go-scm lib, hence listBranchesV2 and once everyone are on board with the ListBranchesV2, then gradually we can deprecate the listBranches method

PR #261, opened 2023-07-12
[feat]: [CDS-73572]: Support List Repo Live Search for all git providers

adivishy1:
Have added code for supporting text based repo filtration for the following git providers:
1. Bitbucket cloud
2. Bitbucket server
3. GitHub
4. Gitlab

Note: Azure does not provide support for text based search.
Have tested out string based search all the above providers. For the other drivers that implement the GitService interface, have called the ListRepo method which means that its not filtering based on any repo being passed.


scm/client.go
```diff
@@ -75,6 +75,14 @@ type (
 		PageListOptions ListOptions
 	}
 
+	// RepoListOptions specifies optional repo search term and pagination
+	// parameters.
+	RepoListOptions struct {
+		SearchTerm      string
+		User            string
+		PageListOptions ListOptions
+	}
```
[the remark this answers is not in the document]
adivishy1: Thanks for this tip, wasnt aware about it, have incorporated the suggested change

abhinav-harness:
Any chance you can also make it work for harness scm ?

adivishy1:
> Any chance you can also make it work for harness scm ?

At present we wanted to implement for GIt providers mentioned in the pr, we can take this enhancement for harness scm separately

adivishy1:
> I noticed some files were added that use camelcase, for example:
> 
> ```
> scm/driver/bitbucket/testdata/reposFilter.json
> ```
> 
> This repository uses snake case (e.g. `repos_filter.json`). To ensure consistency with the rest of the repository, can you please update any files that use camel case to snake case? Thanks!

Done

adivishy1:
> I noticed some files were added that use camelcase, for example:
> 
> ```
> scm/driver/bitbucket/testdata/reposFilter.json
> ```
> 
> This repository uses snake case (e.g. `repos_filter.json`). To ensure consistency with the rest of the repository, can you please update any files that use camel case to snake case? Thanks!

Have incorporated the same changes to my previous pr for supporting live search for listing branches, hence few more file changes are appearing as part of this pr

tphoney (changes requested):
Can you make sure you are consistent adding comments for the V2 functionality if that SCM provider doesnt support search filters. 
This also raises the question that when 3rd party developers that use this V2 functionality will have no idea whether the filters are working for all SCM providers, as they will always get a list result.  But it wont be filtered.
I am not 100% sure on whether returning 
` return nil, nil, scm.ErrNotSupported` is a better solution. @bradrydzewski thoughts ?

scm/driver/gitea/repo.go
```diff
@@ -46,6 +46,10 @@ func (s *repositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*
 	return convertRepositoryList(out), res, err
 }
 
+func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions) ([]*scm.Repository, *scm.Response, error) {
+	return s.List(ctx, opts.ListOptions)
```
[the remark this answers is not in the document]
adivishy1: done

scm/driver/gitee/repo.go
```diff
@@ -45,6 +45,9 @@ func (s *RepositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*
 	res, err := s.client.do(ctx, "GET", path, nil, &out)
 	return convertRepositoryList(out), res, err
 }
+func (s *RepositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions) ([]*scm.Repository, *scm.Response, error) {
+	return s.List(ctx, opts.ListOptions)
```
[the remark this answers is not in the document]
adivishy1: done

adivishy1:
> Can you make sure you are consistent adding comments for the V2 functionality if that SCM provider doesnt support search filters. This also raises the question that when 3rd party developers that use this V2 functionality will have no idea whether the filters are working for all SCM providers, as they will always get a list result. But it wont be filtered. I am not 100% sure on whether returning ` return nil, nil, scm.ErrNotSupported` is a better solution. @bradrydzewski thoughts ?

I have added comments as to why we are calling the List api from V2.

For the 3rd party developers, they should continue to use the List api itself. IMO they should be knowing if the git provider that they are using provides search functionality or not and based on that they can update to the ListV2 apis. We can update our Harness docs as by calling out which all providers are supporting search filters.

If we go by `return nil, nil, scm.ErrNotSupported` , our scm service code will become ugly in which we will have to write code that will branch based on git providers to either call List or ListV2 apis. 

As per the current implementation in this pr, the code will be flexible, in the scm service we just call the ListV2 api, we just add search term if passed or not. For the git providers that dont support search term, we return the complete list which will be in line with the existing prod behaviour, this will give us similar response as calling the List api

scm/client.go
```diff
@@ -71,8 +71,21 @@ type (
 	// BranchListOptions specifies optional branch search term and pagination
 	// parameters.
 	BranchListOptions struct {
-		SearchTerm      string
-		PageListOptions ListOptions
+		ListOptions
```
mohitg0795, line 74: Changing this struct would be a backward incompatible change.

adivishy1: agreed, have reverted this

scm/driver/gitlab/repo.go
```diff
@@ -94,6 +94,13 @@ func (s *repositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*
 	return convertRepositoryList(out), res, err
 }
 
+func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions) ([]*scm.Repository, *scm.Response, error) {
```
raghavharness, line 97: Add a comment explaining this API

scm/driver/github/util.go
```diff
@@ -22,6 +23,33 @@ func encodeListOptions(opts scm.ListOptions) string {
 	return params.Encode()
 }
 
+func formRepoListQueryParams(opts scm.RepoListOptions) string {
```
raghavharness, line 26: Can we keep the name consistent to `encodeRepoListOptions` for each provider?

scm/driver/github/repo.go
```diff
@@ -110,6 +114,14 @@ func (s *RepositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*
 	return convertRepositoryList(out), res, err
 }
 
+// ListV2 returns the user repository list based on the searchTerm passed.
+func (s *RepositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions) ([]*scm.Repository, *scm.Response, error) {
+	path := fmt.Sprintf("search/repositories?%s", formRepoListQueryParams(opts))
+	out := new(searchRepositoryList)
+	res, err := s.client.do(ctx, "GET", path, nil, &out)
+	return convertSearchRepositoryList(out), res, err
```
raghavharness, line 122: We can use the existing `convertRepositoryList` method here ? just pass `out.Repositories`

raghavharness (changes requested):

scm/driver/gitlab/repo.go
adivishy1: done

scm/driver/github/util.go
adivishy1: For github we are not encoding the query params like we are doing for the other providers, hence different name.
When we encode the query params and try the same out in the curls and through the code, it doesnt work, hence no encoding query params for github

scm/driver/github/repo.go
adivishy1: done

scm/driver/github/util.go
raghavharness: Okay can you once check for gitlab ?

adivishy1: for gitlab, in the existing List api, we are calling, `encodeMemberListOptions` hence in the ListV2 api we are calling `encodeMemberRepoListOptions` keeping the naming inline with the existing names

raghavharness (approved):

mohitg0795 (approved):

PR #265, opened 2023-08-14
[IAC-941]: PR comment creation for BitBucket

scottyw-harness:
BitBucket doesn't implement support for PR comments or issue comments.

The current implementation of "pullService" in BitBucket defers to the implementation of "issueService", which itself returns a "not implemented" error for all comment-related calls.

PR comments and issue comments are separate in BitBucket so this commit adds explicit implementations of the comment functions to the "pullService". Most remain "not implemented" with the exception of comment creation which is now supported.


scottyw-harness:
Verified against BitBucket:

image

scottyw-harness:
Verified against BitBucket:
image

PR #270, opened 2023-08-29
[feat]: [CDS-75848]: Add new action type for github provider

rathodmeetsatish:


tphoney (changes requested):
The code implementation looks good and it has unit tests as well. 
My issue would be, do other SCM providers have this concept ?
The whole point of go-scm is to abstract away implementation details from each SCM to give a generic way of dealing with SCM systems as a whole. 
If this is only a github only feature, it does not make sense to add this as a new actiontype. 

So can you look to see if this concept of pr approval exists in the other supported SCM webhooks. eg gitlab / gogs / bitbucket

scm/driver/gitlab/webhook.go
```diff
@@ -338,6 +338,9 @@ func convertPullRequestHook(src *pullRequestHook) *scm.PullRequestHook {
 		action = scm.ActionMerge
 	case "update":
 		action = scm.ActionSync
+		if src.Changes.Draft.Previous.Bool == false {
```
rathodmeetsatish, line 341: @bradrydzewski should this be src.Changes.Draft.Previous.Bool == false && src.Changes.Draft.Current.Bool == true

tphoney: there seems to be a number of values available for the previous field. that isnt a boolean. eg https://github.com/jenkinsci/gitlab-plugin/issues/770 previous can be locked. 
In the gitlab ui, you can inspect/test the webhooks being sent. and specifically around ready for review
eg using ngrok
image

bradrydzewski: @tphoney only the `changes.draft` field is mapped into the struct

```
		Changes struct {
			Draft struct {
				Previous null.Bool `json:"previous"`
				Current  null.Bool `json:"current"`
			} `json:"draft"`
		} `json:"changes"`
```

so when the below json is unmarshaled into the above `Changes` struct, the `draft` field is unmarshaled into `Draft`.  The `title`, `updated_at` and `updated_by_id` fields would be ignored because there have no associated structs in `Changes`.

```
  "changes": {
    "draft": {
      "previous": false,
      "current": true
    },
    "title": {
      "previous": "Update README.md",
      "current": "Draft: Update README.md"
    },
    "updated_at": {
      "previous": "2023-09-07 19:42:35 UTC",
      "current": "2023-09-07 19:42:45 UTC"
    },
    "updated_by_id": {
      "previous": null,
      "current": 13900456
    }
  },
```

so I think we are good, but let me know if there is something else we need to consider

tphoney (approved):

PR #285, opened 2023-12-07
fix: change login to display name for harness

abhinav-harness:


bradrydzewski:
hey @abhinav-harness the `login` parameter should be the unique identifier for the user.  The consumers of this library will rely on that being the unique identifier, and so I think using display names will cause problems for consumers of the library.

bradrydzewski (changes requested):
`login` must use a unique identifier / unique key in the gitness database as the value (user identifier or email will suffice)

abhinav-harness:
this was needed by ci service as consumer to display harness user names but will discard this PR and handle it on ci side.

PR #286, opened 2023-12-08
fix: use opts for harness list commits

abhinav-harness:


devkimittal (approved):
Assuming this has been tested e2e

PR #289, opened 2023-12-20
feat: support more events in webhook parse in go-scm for gitness

abhinav-harness:


scm/driver/harness/webhook.go
```diff
@@ -271,6 +275,7 @@ func convertPullReq(pr pullReq, ref ref, commit hookCommit) scm.PullRequest {
 		Closed: pr.State != "open",
```
rutvijmehta-harness, line 275: Do we need `pr.State == "closed"`?

johannesHarness: agree

scm/driver/harness/webhook.go
```diff
@@ -271,6 +275,7 @@ func convertPullReq(pr pullReq, ref ref, commit hookCommit) scm.PullRequest {
...
 		Source: pr.SourceBranch,
 		Target: pr.TargetBranch,
+		Merged: pr.State == "merged",
 		Fork:   "fork",
```
johannesHarness, line 279: should we keep this empty? not sure why this says "fork"

scm/driver/harness/webhook.go
abhinav-harness: pr merged is also pr closed, that is what we are doing in github's case too.

scm/driver/harness/webhook.go
abhinav-harness: don't want to change something which is working in this PR ⛔

scm/driver/harness/webhook.go
abhinav-harness: can change it.

johannesHarness (approved):

rutvijmehta-harness (approved):

PR #301, opened 2024-03-19
[feat]: [CDS-93606]: Added Support for search term in For List Repo Operations.

senjucanon2:
## Issue:
- Github support search term in For List Repo Operations but our flow was failing for the same. Issue was pinpointes at the go-scm library level.

## Changes:
- Added Blank Space in Url Encoded Format to build the correct Url path.

## Testing:
- Changes were tested for various search terms on my Github Account with user name "senjucanon2": 
- - Search term "test" -> fetched 3 results: "test-repo", "test-repo2", "terraform-test".
- - Search term "test-repo" -> fetched 2 results: "test-repo", "test-repo2"
- - Search term "terraform" -> fetched 1 result: "terraform-test".
- - Invalid Search term -> fetched 0 result.

### DIFF
--- a/scm/driver/github/util.go
+++ b/scm/driver/github/util.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"fmt"
 
 	"github.com/drone/go-scm/scm"
 )
--- a/scm/driver/github/util.go
+++ b/scm/driver/github/util.go
@@ -29,7 +29,7 @@ func encodeRepoListOptions(opts scm.RepoListOptions) string {
 		if opts.RepoSearchTerm.RepoName != "" {
 			sb.WriteString("q=")
 			sb.WriteString(opts.RepoSearchTerm.RepoName)
-			sb.WriteString("in:name+user:")
+			sb.WriteString(" in:name+user:")
 			sb.WriteString(opts.RepoSearchTerm.User)
 		} else {
 			sb.WriteString("q=")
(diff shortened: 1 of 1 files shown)


bradrydzewski:
I noticed this function uses a `strings.Builder` for building the url query string. However, in the rest of the functions we use `url.Values` for building and encoding the query string.  Example:

```Go
	params := url.Values{}
	if opts.Page != 0 {
		params.Set("page", strconv.Itoa(opts.Page))
	}
	if opts.Size != 0 {
		params.Set("per_page", strconv.Itoa(opts.Size))
	}
	...
	return params.Encode()
```

Is there a reason to use `strings.Builder` or can we change this to `url.Values`?

scm/driver/github/util.go
```diff
@@ -29,7 +29,8 @@ func encodeRepoListOptions(opts scm.RepoListOptions) string {
 		if opts.RepoSearchTerm.RepoName != "" {
 			sb.WriteString("q=")
 			sb.WriteString(opts.RepoSearchTerm.RepoName)
-			sb.WriteString("in:name+user:")
+			// %20 is urlEncoding of blank space.
+			sb.WriteString("%20in:name+user:")
```
bradrydzewski, line 33: can we use `url.Values` to encode instead of string builder?

bradrydzewski (changes requested):

scm/driver/github/util.go
senjucanon2: @bradrydzewski  
We have to use the string builder here as we need to make manually make a string for query at Github Side. Here we are not adding query params in the API. 

Here's the link for reference in the API we are using : https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28.
Thanks for the insightful review.

adivishy1 (approved):

mohitg0795 (approved):

bradrydzewski (approved):

PR #302, opened 2024-03-20
feat: pr harness parse all fields

abhinav-harness:


scm/driver/harness/pr.go
```diff
@@ -234,9 +229,21 @@ func convertPullRequest(src *pr) *scm.PullRequest {
...
-		Closed: src.State == "closed",
+		Head: scm.Reference{
+			Name: src.SourceBranch,
+			Path: scm.ExpandRef(src.SourceBranch, "refs/heads"),
+			Sha:  src.SourceSHA,
+		},
+		Base: scm.Reference{
+			Name: src.TargetBranch,
+			Path: scm.ExpandRef(src.TargetBranch, "refs/heads"),
+			Sha:  src.MergeSHA,
```
abhinav-harness, line 240: should it be merge sha or mergebasesha?

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
+		SourceSHA    string `json:"source_sha"`
 		TargetBranch string `json:"target_branch"`
 		TargetRepoID int    `json:"target_repo_id"`
-		Title        string `json:"title"`
+
+		State string `json:"state"`
+		Stats struct {
+			Commits         int `json:"commits"`
+			Conversations   int `json:"conversations"`
+			FilesChanged    int `json:"files_changed"`
```
johannesHarness, line 129: all entries in here have omitempty

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
-		Number       int    `json:"number"`
+		Author      principal `json:"author"`
+		Created     int64     `json:"created"`
+		Description string    `json:"description"`
+		Edited      int64     `json:"edited"`
+		IsDraft     bool      `json:"is_draft"`
+
+		MergeTargetSHA   string    `json:"merge_target_sha"`
+		MergeBaseSha     string    `json:"merge_base_sha"`
+		Merged           int       `json:"merged"`
```
johannesHarness, line 110: should be `*int64`

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
-			Updated     int    `json:"updated"`
-		} `json:"merger"`
-		Number       int    `json:"number"`
+		Author      principal `json:"author"`
+		Created     int64     `json:"created"`
+		Description string    `json:"description"`
+		Edited      int64     `json:"edited"`
+		IsDraft     bool      `json:"is_draft"`
+
+		MergeTargetSHA   string    `json:"merge_target_sha"`
```
johannesHarness, line 108: is `*string`

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
+		MergeBaseSha     string    `json:"merge_base_sha"`
+		Merged           int       `json:"merged"`
+		MergeMethod      string    `json:"merge_method"`
+		MergeSHA         string    `json:"merge_sha"`
```
johannesHarness, line 112: is `*string`

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
+		MergeCheckStatus string    `json:"merge_check_status"`
+		MergeConflicts   []string  `json:"merge_conflicts,omitempty"`
+		Merger           principal `json:"merger"`
+
+		Number int `json:"number"`
+
 		SourceBranch string `json:"source_branch"`
 		SourceRepoID int    `json:"source_repo_id"`
```
johannesHarness, line 120: int64

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
-		State        string `json:"state"`
-		Stats        struct {
-			Commits       int `json:"commits"`
-			Conversations int `json:"conversations"`
-			FilesChanged  int `json:"files_changed"`
-		} `json:"stats"`
+		SourceSHA    string `json:"source_sha"`
 		TargetBranch string `json:"target_branch"`
 		TargetRepoID int    `json:"target_repo_id"`
```
johannesHarness, line 123: int64

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
-		Title        string `json:"title"`
+
+		State string `json:"state"`
+		Stats struct {
+			Commits         int `json:"commits"`
+			Conversations   int `json:"conversations"`
```
johannesHarness, line 128: both int64 (and technically `*int64`)

scm/driver/harness/pr.go
abhinav-harness: cant make this change as base struct only supports int

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
+		IsDraft     bool      `json:"is_draft"`
+
+		MergeTargetSHA   string    `json:"merge_target_sha"`
+		MergeBaseSha     string    `json:"merge_base_sha"`
+		Merged           int       `json:"merged"`
+		MergeMethod      string    `json:"merge_method"`
+		MergeSHA         string    `json:"merge_sha"`
+		MergeCheckStatus string    `json:"merge_check_status"`
+		MergeConflicts   []string  `json:"merge_conflicts,omitempty"`
+		Merger           principal `json:"merger"`
```
johannesHarness, line 115: `*principal`

scm/driver/harness/pr.go
abhinav-harness: would pointer make any difference in json response object?

scm/driver/harness/pr.go
johannesHarness: I assume that's the sha of the target branch? if that's the case then it should be `MergeTargetSHA` - that's the SHA of the target branch when the last merge check was performed (if no merge check was performed - either it didn't happen yet or target was updated and it was reset - this value is nil btw)

scm/driver/harness/pr.go
```diff
@@ -234,9 +229,21 @@ func convertPullRequest(src *pr) *scm.PullRequest {
...
+			Name: src.SourceBranch,
+			Path: scm.ExpandRef(src.SourceBranch, "refs/heads"),
+			Sha:  src.SourceSHA,
+		},
+		Base: scm.Reference{
+			Name: src.TargetBranch,
+			Path: scm.ExpandRef(src.TargetBranch, "refs/heads"),
+			Sha:  src.MergeSHA,
+		},
+		Fork:    "fork",
```
johannesHarness, line 242: should we set fork to anything - I'd assume it should be empty, no?

abhinav-harness: i think empty string can cause some issue for upper grpc consumer

scm/driver/harness/pr.go
johannesHarness: I think we should still have int64 here to stay true - and then when converting it we cast it.
Otherwise it's very implicit and people might not notice

scm/driver/harness/pr.go
johannesHarness: not much - but it's an optional value - and also I think we should stay true to the original if possible?

scm/driver/harness/pr.go
johannesHarness: oh okay we can keep it - seems odd though

scm/driver/harness/pr.go
abhinav-harness: what is MergeBaseSha - i think it should be `mergebasesha` in that case - i saw github they are using base.sha but i dont know what the value points to

abhinav-harness: probably the head sha when PR was raised?

johannesHarness: >probably the head sha when PR was raised?

I don't think we can use that, it'll  be outdated quite soon.
We should use what I mentioned.

`mergebasesha` is the sha that was the biggest common ancestor commit between target and source branch that was used at time of mergecheck.
We shouldn't use that either - could be far in the past if it's an older PR that wasn't rebased.

As I said, if we want the latest sha of target branch that was used for mergecheck then it's MergeTargetSHA.
If we want latest latest target sha - that we don't have here.

scm/driver/harness/pr.go
```diff
@@ -99,44 +99,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
-		Number       int    `json:"number"`
+		Author      principal `json:"author"`
+		Created     int64     `json:"created"`
+		Description string    `json:"description"`
+		Edited      int64     `json:"edited"`
+		IsDraft     bool      `json:"is_draft"`
+
+		MergeTargetSHA   *string    `json:"merge_target_sha"`
+		MergeBaseSha     string     `json:"merge_base_sha"`
+		Merged           *int64     `json:"merged"`
```
bradrydzewski, line 110: for consistency with other driver implementations you can use `null.Int64` and `null.String`.  example here https://github.com/drone/go-scm/blob/fcc8453fa15d03845c2017d0388d9e1e159d0831/scm/driver/gitlab/org.go#L40

bradrydzewski (changes requested):

scm/driver/harness/pr.go
```diff
@@ -99,44 +100,38 @@ func (s *pullService) Close(context.Context, string, int) (*scm.Response, error)
...
+		Author      principal `json:"author"`
+		Created     int64     `json:"created"`
+		Description string    `json:"description"`
+		Edited      int64     `json:"edited"`
+		IsDraft     bool      `json:"is_draft"`
+
+		MergeTargetSHA   null.String `json:"merge_target_sha"`
+		MergeBaseSha     string      `json:"merge_base_sha"`
+		Merged           null.Int    `json:"merged"`
+		MergeMethod      string      `json:"merge_method"`
```
johannesHarness, line 112: this is also a `null.String` (technically an enum)

johannesHarness (approved):

johannesHarness (approved):

PR #306, opened 2024-05-08
[feat]: [CI-12179]: Add support for gitlab basesha for PRs

smjt-h:


ShobhitSingh11 (commented):
Code seems okay, wondering if we are supporting basesha then what will be use of headsha ans startsha

smjt-h:
> Code seems okay, wondering if we are supporting basesha then what will be use of headsha ans startsha

Base sha and Start sha are same, head sha is source branch sha, we are populating base sha as of now because of customer requirement. They want to use this variable in CI for gitlab PR builds

ShobhitSingh11 (approved):

PR #309, opened 2024-05-29
feat: add list namespace for repos

abhinav-harness:


scm/driver/gitea/repo.go
```diff
@@ -51,6 +51,11 @@ func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions
 	return s.List(ctx, opts.ListOptions)
 }
 
+func (s *repositoryService) ListNamespace(ctx context.Context, _ string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	// gitea does not have namespace
+	return s.List(ctx, opts)
```
johannesHarness, line 56: gitea has organizations just like github, no?

johannesHarness: If there's no API then maybe we should return an error instead of returning wrong data?

scm/driver/gitlab/repo.go
```diff
@@ -102,6 +102,10 @@ func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions
 	return convertRepositoryList(out), res, err
 }
 
+func (s *repositoryService) ListNamespace(ctx context.Context, namespace string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	return nil, nil, scm.ErrNotSupported
```
johannesHarness, line 106: gitlab should have an api to get projects in a group path:
https://docs.gitlab.com/ee/api/groups.html#list-a-groups-projects

scm/driver/harness/repo.go
```diff
@@ -68,6 +68,11 @@ func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions
 	return s.List(ctx, opts.ListOptions)
 }
 
+func (s *repositoryService) ListNamespace(ctx context.Context, _ string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	// List already has context about namespace
```
johannesHarness, line 72: **minor** client has context, correcr?

johannesHarness: *random thought* I wonder whether we should change that at some point? at least only keep the accountid on client level?

scm/driver/stash/repo.go
```diff
@@ -192,6 +192,18 @@ func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions
...
 }
 
+// ListNamespace returns the user repository list based on searchterm and namespace.
+func (s *repositoryService) ListNamespace(ctx context.Context, namespace string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	path := fmt.Sprintf("rest/api/1.0/projects/%s/repos?%s", namespace, encodeListRoleOptions(opts))
+	out := new(repositories)
+	res, err := s.client.do(ctx, "GET", path, nil, &out)
+	if res != nil && !out.pagination.LastPage.Bool {
+		res.Page.First = 1
+		res.Page.Next = opts.Page + 1
```
johannesHarness, line 202: this differes from the above, should be `opts.ListOptions.Page + 1`?

abhinav-harness: @johannesHarness this is copied from List call. ListV2 has addition of search functionality which doesn't make sense here.

scm/driver/harness/repo.go
abhinav-harness: done, and later maybe

scm/driver/gitlab/repo.go
abhinav-harness: done

scm/driver/gitea/repo.go
abhinav-harness: fixed

scm/driver/gitlab/repo.go
```diff
@@ -102,6 +102,13 @@ func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions
 	return convertRepositoryList(out), res, err
 }
 
+func (s *repositoryService) ListNamespace(ctx context.Context, namespace string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
+	path := fmt.Sprintf("api/v4//groups/%s/projects?%s", namespace, encodeMemberListOptions(opts))
```
johannesHarness, line 106: double `//`

scm/driver/stash/repo.go
johannesHarness: ohh didn't expand that much. makes sense, thanks

PR #310, opened 2024-05-30
feat: [IAC-1930]: added ListTags

vitsafronovici:


scm/driver/bitbucket/git.go
```diff
@@ -81,7 +81,7 @@ func (s *gitService) ListCommits(ctx context.Context, repo string, opts scm.Comm
 }
 
 func (s *gitService) ListTags(ctx context.Context, repo string, opts scm.ListOptions) ([]*scm.Reference, *scm.Response, error) {
-	path := fmt.Sprintf("2.0/repositories/%s/refs/tags?%s", repo, encodeListOptions(opts))
```
bradrydzewski, line 84: `encodeListOptions` is typically required for pagination and should not be removed

bradrydzewski (changes requested):

scm/driver/bitbucket/git.go
bradrydzewski: this works for me:

```
curl "https://api.bitbucket.org/2.0/repositories/atlassian/atlaskit/refs/tags?pagelen=5"
```

bradrydzewski: I believe you can instead set `Next` to `0` in the response which should force any clients using this library to fallback to the `NextURL` instead for pagination.

```diff
	path := fmt.Sprintf("2.0/repositories/%s/refs/tags?%s", repo, encodeListOptions(opts))
	out := new(branches)
	res, err := s.client.do(ctx, "GET", path, nil, &out)
	copyPagination(out.pagination, res)
+	out.Pagination.Next = 0
	return convertTagList(out), res, err
```

scm/driver/github/git.go
```diff
@@ -191,10 +191,10 @@ func convertRef(from *ref) *scm.Reference {
 	}
 }
 
-func convertTagList(from []*branch) []*scm.Reference {
+func convertTagList(from []*ref) []*scm.Reference {
```
bradrydzewski, line 194: rename to `convertRefList` now that it is converting a slice of refs

bradrydzewski (changes requested):

PR #311, opened 2024-06-06
Fetch PR comments for stash provider

karansaraswat19:


scm/driver/stash/pr.go
```diff
@@ -29,7 +29,7 @@ func (s *pullService) FindComment(ctx context.Context, repo string, number int,
 	path := fmt.Sprintf("rest/api/1.0/projects/%s/repos/%s/pull-requests/%d/comments/%d", namespace, name, number, id)
 	out := new(pullRequestComment)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
-	return convertPullRequestComment(out), res, err
+	return convertPullRequestComment(*out, 0, nil), res, err
```
bradrydzewski, line 32: doing a pointer dereference seems a bit odd here

scm/pr.go
```diff
@@ -9,6 +9,12 @@ import (
 	"time"
 )
 
+const (
+	GeneralCommentType = "comment"
+	CodeCommentType    = "code-comment"
+	ReplyCommentType   = "reply-comment"
+)
```
bradrydzewski, line 16: enums are stored in a `const.go` file

bradrydzewski:
let's scope this pull request to issue comments only, and handle code comments in a separate PR.  as discussed in slack, we may want code comments to be a separate method and separate set of struct, but in this PR it looks like they are being combined.

scm/pr.go
```diff
@@ -61,6 +67,20 @@ type (
 		PrevFilePath string
 	}
 
+	GeneralCommentMetadata struct{}
+
+	ReplyCommentMetadata struct {
+		ParentID int
```
bradrydzewski, line 73: keep in mind not every provider uses a numerical identifier, which is why it is important to research all providers when designing a new abstraction

scm/issue.go
```diff
@@ -43,11 +43,13 @@ type (
...
-		Author  User
-		Created time.Time
-		Updated time.Time
+		ID       int
+		Body     string
+		Author   User
+		Created  time.Time
+		Updated  time.Time
+		Type     string
+		Metadata interface{}
```
bradrydzewski, line 52: I recommend we remove `type` and `metadata` at this time until we have a design for all providers (github, gitlab, etc).

bradrydzewski:
> Will make another method ListPRCodeComments after all the research.

Sounds great.  Note that we can call it `ListCodeComments`

```diff
-ListPRCodeComments
+ListCodeComments
```

The reason is that this method is part of `PullRequestService`.  We know that `ListCodeComments` is listing pull request code comments because it is a member of the pull request service, so no need to include `PR` in the method name.

This document can be a good reference for naming conventions in Go:
https://google.github.io/styleguide/go/decisions#external-context-vs-local-names

karansaraswat19:
> > Will make another method ListPRCodeComments after all the research.
> 
> Sounds great. Note that we can call it `ListCodeComments`
> 
> ```diff
> -ListPRCodeComments
> +ListCodeComments
> ```
> 
> The reason is that this method is part of `PullRequestService`. We know that `ListCodeComments` is listing pull request code comments because it is a member of the pull request service, so no need to include `PR` in the method name.
> 
> This document can be a good reference for naming conventions in Go: https://google.github.io/styleguide/go/decisions#external-context-vs-local-names

Thanks @bradrydzewski! Got it.
New to `Go`, getting to know new new things everyday
`should be a private type (lowercase)`
Didn't know about this also earlier

scm/pr.go
karansaraswat19: removed for now

scm/pr.go
karansaraswat19: done

scm/issue.go
karansaraswat19: done

scm/driver/stash/pr.go
karansaraswat19: removed

PR #312, opened 2024-06-17
fix query params for List PRs API call for stash provider for paginat…

karansaraswat19:
…ion and state of PRs

### DIFF
--- a/scm/driver/stash/util.go
+++ b/scm/driver/stash/util.go
@@ -111,7 +111,7 @@ func encodePullRequestListOptions(opts scm.PullRequestListOptions) string {
 	if opts.Open && opts.Closed {
 		params.Set("state", "all")
 	} else if opts.Closed {
-		params.Set("state", "closed")
+		params.Set("state", "declined")
 	}
 	return params.Encode()
 }
--- a/scm/driver/stash/pr.go
+++ b/scm/driver/stash/pr.go
@@ -34,7 +34,7 @@ func (s *pullService) FindComment(ctx context.Context, repo string, number int, 
 
 func (s *pullService) List(ctx context.Context, repo string, opts scm.PullRequestListOptions) ([]*scm.PullRequest, *scm.Response, error) {
 	namespace, name := scm.Split(repo)
-	path := fmt.Sprintf("rest/api/1.0/projects/%s/repos/%s/pull-requests", namespace, name)
+	path := fmt.Sprintf("rest/api/1.0/projects/%s/repos/%s/pull-requests?%s", namespace, name, encodePullRequestListOptions(opts))
 	out := new(prs)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
 	if !out.pagination.LastPage.Bool {


scm/pr.go
```diff
@@ -48,6 +48,7 @@ type (
 		Size   int
 		Open   bool
 		Closed bool
+		Merged bool
```
abhinav-harness, line 51: this is added in a common file, we would need addition for all providers for this if they support

scm/driver/stash/util.go
```diff
@@ -108,10 +108,12 @@ func encodePullRequestListOptions(opts scm.PullRequestListOptions) string {
 	if opts.Size != 0 {
 		params.Set("limit", strconv.Itoa(opts.Size))
 	}
-	if opts.Open && opts.Closed {
+	if opts.Open && opts.Closed && opts.Merged {
```
abhinav-harness, line 111: of what i understand we need state all -> lets not make too many changes and in caller add both open and close

scm/pr.go
karansaraswat19: Gitlab now has 4 states -> `opened, closed, locked, merged`
Bitbucket also now has 4 -> `OPEN, MERGED, DECLINED, SUPERSEDED`
Till now everywhere we have this condition though
`if opts.Open && opts.Closed {
	params.Set("state", "all")
}`

scm/driver/stash/util.go
karansaraswat19: did this according to the changes already present with this condition
`if opts.Open && opts.Closed { params.Set("state", "all") }`

scm/pr.go
karansaraswat19: this might affect the usage of this method

bradrydzewski (changes requested):
When adding new functionality to this library at the abstraction layer (in this case `scm.PullRequest.Merged`) we need to add support for multiple vendors (e.g. GitHub, GItLab and Stash) to ensure the abstraction is correct.

scm/pr.go
bradrydzewski: one requirement to merging new functionality in this library is that it needs to be implemented for 3 or more providers to ensure the abstractions are correct.  The differences you highlighted are the reason this is required.

karansaraswat19: removed

scm/driver/stash/util.go
karansaraswat19: fixed

karansaraswat19:
> When adding new functionality to this library at the abstraction layer (in this case `scm.PullRequest.Merged`) we need to add support for multiple vendors (e.g. GitHub, GItLab and Stash) to ensure the abstraction is correct.

Noted

abhinav-harness (approved):

scm/driver/stash/util.go
```diff
@@ -111,7 +111,7 @@ func encodePullRequestListOptions(opts scm.PullRequestListOptions) string {
 	if opts.Open && opts.Closed {
 		params.Set("state", "all")
 	} else if opts.Closed {
-		params.Set("state", "closed")
+		params.Set("state", "declined")
```
karansaraswat19, line 114: stash calls its closed PRs as "declined" PRs
```
"errors": [
        {
            "context": null,
            "message": "Pull request state should be one of: [DECLINED, MERGED, OPEN]",
            "exceptionName": null
        }
    ]
    ```

PR #318, opened 2024-08-02
feat: [CODE-2180]: Add draft boolean for github, azure and gitlab PRs

karansaraswat19:


scm/driver/azure/pr.go
```diff
@@ -107,6 +107,7 @@ type pr struct {
 	SourceRefName         string      `json:"sourceRefName"`
 	TargetRefName         string      `json:"targetRefName"`
 	MergeStatus           string      `json:"mergeStatus"`
+	IsDraft               bool        `json:"isDraft"`
```
abhinav-harness, line 110: can we call it draft at all places and just rename json names?
I see we follow that convention - if not we can keep isDraft.

karansaraswat19: I read somewhere that the variable name need to start with uppercase letter if we expect it to be marshalled

karansaraswat19:
> can you also check bitbucket cloud

> As of now, Bitbucket Cloud does not support draft pull requests. However, this feature is available in other major platforms as listed above.
They still don't support it

scm/driver/azure/pr.go
abhinav-harness: I meant something - but nvm - we are good

abhinav-harness (approved):

PR #322, opened 2024-08-29
fix: pass gitlab mergerequest param via json body

olblak:
When opening a merge request on GitLab, the current implementation pass all the parameters via the URL parameters instead of using the HTTP body.
The problem is that the amount of information that can be provided via URL parameter is very limited, which lead to not being able to create a merge request.

## Test

On top of the unit test, I tested this change on a different project:

<details><summary>Screenshot</summary>

image

</details> 

### Potential improvement

I also noticed that function `⁣CreateComment` use URL parameter and I think it could also be improved but I am not using that function on my project

### DIFF
--- a/scm/driver/gitlab/pr.go
+++ b/scm/driver/gitlab/pr.go
@@ -60,14 +60,23 @@ func (s *pullService) ListCommits(ctx context.Context, repo string, number int, 
 }
 
 func (s *pullService) Create(ctx context.Context, repo string, input *scm.PullRequestInput) (*scm.PullRequest, *scm.Response, error) {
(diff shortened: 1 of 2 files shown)


olblak:
@abhinav-harness could you take a look at this pull request, as it fix a bug with the GitLab merge request creation

abhinav-harness (approved):

olblak:
@abhinav-harness Would it be possible to have this pullrequest available in a futur release?

abhinav-harness:
@olblak released https://github.com/drone/go-scm/releases/tag/v1.39.0

PR #326, opened 2024-11-19
Update Bitbucket PR author login

atefehmohseni:
Currently Bitbucket PR author login field uses user's nickname which is not unique. I suggest to use accountID or UUID instead as explained here:
https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-changes-gdpr/

### DIFF
--- a/scm/driver/bitbucket/pr.go
+++ b/scm/driver/bitbucket/pr.go
@@ -193,7 +193,7 @@ func convertPullRequest(from *pr) *scm.PullRequest {
 			Sha:  from.Destination.Commit.Hash,
 		},
 		Author: scm.User{
-			ID:  from.Author.AccountID,
+			ID:     from.Author.AccountID,
 			Name:   from.Author.DisplayName,
 			Avatar: from.Author.Links.Avatar.Href,
 		},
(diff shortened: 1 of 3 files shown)


bradrydzewski (changes requested):
We have an `ID` field that is meant to hold the unique identifier.  Let's use this instead for the author account id.

```go
	User struct {
		ID      string
		Login   string
		Name    string
		Email   string
		Avatar  string
		Created time.Time
		Updated time.Time
	}
```

```diff
		Author: scm.User{
+			ID:     from.Author.AccountID,
			Login:  from.Author.Nickname,
			Name:   from.Author.DisplayName,
			Avatar: from.Author.Links.Avatar.Href,
		},
```

**EDIT** as an aside, the `Login` is currently being set to the Nickname.  The `Login` field should only be set to the username, and if no username is available, it should be empty.  We should probably revert this because it will result in a bug for anyone that uses it.

```
		Author: scm.User{
-			Login:  from.Author.Nickname,
			Name:   from.Author.DisplayName,
			Avatar: from.Author.Links.Avatar.Href,
		},
```

atefehmohseni:
Done.

scm/driver/bitbucket/pr.go
```diff
@@ -193,7 +193,7 @@ func convertPullRequest(from *pr) *scm.PullRequest {
 			Sha:  from.Destination.Commit.Hash,
 		},
 		Author: scm.User{
-			Login:  from.Author.Nickname,
+			ID:  from.Author.AccountID,
```
bradrydzewski, line 196: run gofmt

atefehmohseni: Sure, done.

PR #351, opened 2026-02-13
feat: [PIPE-32185]: handle deprecated cross-workspace API endpoints

ritek01:
feat: [PIPE-32185]: handle deprecated cross-workspace API endpoints   
   - Replace /2.0/workspaces with /2.0/user/workspaces for org listing
   - Replace /2.0/repositories with workspace-scoped aggregation for repo listing
   - Replace /2.0/user/permissions/repositories with workspace-scoped endpoint
   - Add workspace extraction from client URL with fallback iteration

Testing : 
  - Service is starting & working
  - List Repos is working
  - PR is getting created

Screenshots : 
 
image
image
image

[PIPE-32185]: https://harness.atlassian.net/browse/PIPE-32185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scm/driver/bitbucket/org.go
+++ b/scm/driver/bitbucket/org.go
@@ -11,6 +11,10 @@ import (
 	"github.com/drone/go-scm/scm"
 )
 
+const (
+	avatarURLTemplate = "https://bitbucket.org/account/%s/avatar/32/"
+)
+
 type organizationService struct {
 	client *wrapper
 }
@@ -36,9 +40,9 @@ func (s *organizationService) List(ctx context.Context, opts scm.ListOptions) ([
 
 func convertWorkspaceAccessList(from *workspaceAccessList) []*scm.Organization {
 	to := []*scm.Organization{}
-	for _, values := range from.Values {
-		if values.Workspace != nil {
-			to = append(to, convertWorkspace(values.Workspace))
+	for _, value := range from.Values {
+		if value.Workspace != nil {
+			to = append(to, convertWorkspace(value.Workspace))
 		}
 	}
 	return to
@@ -51,7 +55,7 @@ type organization struct {
 func convertOrganization(from *organization) *scm.Organization {
 	return &scm.Organization{
 		Name:   from.Login,
-		Avatar: fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32/", from.Login),
+		Avatar: fmt.Sprintf(avatarURLTemplate, from.Login),
 	}
 }
 
@@ -60,7 +64,7 @@ func convertWorkspace(workspace *workspace) *scm.Organization {
 	if workspace.Links.Avatar.Href != "" {
 		avatar = workspace.Links.Avatar.Href
 	} else {
-		avatar = fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32", workspace.Slug)
+		avatar = fmt.Sprintf(avatarURLTemplate, workspace.Slug)
 	}
 	return &scm.Organization{
 		Name:   workspace.Slug,
--- a/scm/driver/bitbucket/org.go
+++ b/scm/driver/bitbucket/org.go
@@ -27,24 +27,21 @@ func (s *organizationService) FindMembership(ctx context.Context, name, username
 }
 
 func (s *organizationService) List(ctx context.Context, opts scm.ListOptions) ([]*scm.Organization, *scm.Response, error) {
-	path := fmt.Sprintf("2.0/workspaces?%s", encodeListRoleOptions(opts))
-	out := new(organizationList)
+	path := fmt.Sprintf("2.0/user/workspaces?%s", encodeListRoleOptions(opts))
+	out := new(workspaceAccessList)
 	res, err := s.client.do(ctx, "GET", path, nil, out)
 	copyPagination(out.pagination, res)
-	return convertOrganizationList(out), res, err
+	return convertWorkspaceAccessList(out), res, err
 }
 
-func convertOrganizationList(from *organizationList) []*scm.Organization {
+func convertWorkspaceAccessList(from *workspaceAccessList) []*scm.Organization {
 	to := []*scm.Organization{}
-	for _, v := range from.Values {
-		to = append(to, convertOrganization(v))
+	for _, values := range from.Values {
+		if values.Workspace != nil {
+			to = append(to, convertWorkspace(values.Workspace))
+		}
 	}
 	return to
-}
-
-type organizationList struct {
-	pagination
-	Values []*organization `json:"values"`
 }
 
 type organization struct {
@@ -55,5 +52,18 @@ func convertOrganization(from *organization) *scm.Organization {
 	return &scm.Organization{
 		Name:   from.Login,
 		Avatar: fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32/", from.Login),
+	}
+}
+
+func convertWorkspace(workspace *workspace) *scm.Organization {
+	avatar := ""
+	if workspace.Links.Avatar.Href != "" {
+		avatar = workspace.Links.Avatar.Href
+	} else {
+		avatar = fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32", workspace.Slug)
+	}
+	return &scm.Organization{
+		Name:   workspace.Slug,
+		Avatar: avatar,
 	}
 }
--- a/scm/driver/bitbucket/repo.go
+++ b/scm/driver/bitbucket/repo.go
@@ -134,34 +134,26 @@ func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Pe
 	return nil, nil, fmt.Errorf("repository %s not found in any workspace", repo)
 }
 
-// List returns the user repository list.
+// List returns the user repository list using workspace-aware pagination.
+// Fetches repos across all workspaces (sorted by slug) and fills the page
+// by stitching results from multiple workspaces when needed.
+// Defaults to page 1 if not specified.
 func (s *repositoryService) List(ctx context.Context, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
-	if opts.URL != "" {
-		repoStruct := new(repositories)
-		response, err := s.client.do(ctx, "GET", opts.URL, nil, &repoStruct)
-		if err != nil {
-			return nil, response, err
-		}
-		copyPagination(repoStruct.pagination, response)
-		return convertRepositoryList(repoStruct), response, err
+	if opts.Page == 0 {
+		opts.Page = 1
 	}
-
-	return s.client.fetchReposFromAllWorkspaces(ctx, encodeListRoleOptions(opts))
+	return s.client.fetchReposWithPagination(ctx, encodeListRoleOptions(opts), opts.Page, opts.Size)
 }
 
 // ListV2 returns the user repository list based on the searchTerm passed.
+// Uses the same workspace-aware pagination as List, with an additional
+// search filter applied to each workspace query.
+// Defaults to page 1 if not specified.
 func (s *repositoryService) ListV2(ctx context.Context, opts scm.RepoListOptions) ([]*scm.Repository, *scm.Response, error) {
-	if opts.ListOptions.URL != "" {
-		repoStruct := new(repositories)
-		response, err := s.client.do(ctx, "GET", opts.ListOptions.URL, nil, &repoStruct)
-		if err != nil {
-			return nil, response, err
-		}
-		copyPagination(repoStruct.pagination, response)
-		return convertRepositoryList(repoStruct), response, err
+	if opts.ListOptions.Page == 0 {
+		opts.ListOptions.Page = 1
 	}
-
-	return s.client.fetchReposFromAllWorkspaces(ctx, encodeRepoListOptions(opts))
+	return s.client.fetchReposWithPagination(ctx, encodeRepoListOptions(opts), opts.ListOptions.Page, opts.ListOptions.Size)
 }
 
 func (s *repositoryService) ListNamespace(ctx context.Context, namespace string, opts scm.ListOptions) ([]*scm.Repository, *scm.Response, error) {
--- a/scm/driver/bitbucket/repo.go
+++ b/scm/driver/bitbucket/repo.go
@@ -88,56 +88,71 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
 
 // FindPerms returns the repository permissions.
 func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Perm, *scm.Response, error) {
-	// Try to extract workspace from the client URL
-	workspace := s.client.extractWorkspaceFromURL()
+	// First, try to fetch using the repo identifier provided
+	perm, res, err := s.findPermsWithIdentifier(ctx, repo)
+	if err == nil {
+		return perm, res, nil
+	}
 
-	// If workspace is available, use it to fetch permissions
+	// If that fails and repo isn't in "workspace/repo" format,
+	// search all workspaces as a fallback
+	if !strings.Contains(repo, "/") {
+		return s.findPermsAcrossWorkspaces(ctx, repo)
+	}
+
+	// Repo is in "workspace/repo" format and fetch failed
+	return nil, res, err
+}
+
+// findPermsWithIdentifier attempts to fetch permissions using either the
+// extracted workspace from the client URL or the workspace in "workspace/repo" format.
+func (s *repositoryService) findPermsWithIdentifier(ctx context.Context, repo string) (*scm.Perm, *scm.Response, error) {
+	workspace := s.client.extractWorkspaceFromURL()
+	repoSlug := repo
+
+	// If workspace is in the repo identifier (workspace/repo format), use it
+	if strings.Contains(repo, "/") {
+		workspace, repoSlug = scm.Split(repo)
+	}
+
+	// If we have a workspace (either from URL or repo identifier), fetch permissions
 	if workspace != "" {
-		repoSlug := repo
-		// Handle "workspace/repo" format by extracting just the repo slug
-		if strings.Contains(repo, "/") {
-			_, repoSlug = scm.Split(repo)
-		}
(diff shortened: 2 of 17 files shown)


mohitg0795 (approved):

scm/driver/bitbucket/org.go
```diff
@@ -27,26 +27,23 @@ func (s *organizationService) FindMembership(ctx context.Context, name, username
...
 
-func convertOrganizationList(from *organizationList) []*scm.Organization {
+func convertWorkspaceAccessList(from *workspaceAccessList) []*scm.Organization {
 	to := []*scm.Organization{}
-	for _, v := range from.Values {
-		to = append(to, convertOrganization(v))
+	for _, values := range from.Values {
+		if values.Workspace != nil {
+			to = append(to, convertWorkspace(values.Workspace))
+		}
```
bhumijgupta: nit:
```suggestion
	for _, value := range from.Values {
		if value.Workspace != nil {
			to = append(to, convertWorkspace(value.Workspace))
		}
```

scm/driver/bitbucket/org.go
```diff
@@ -57,3 +54,16 @@ func convertOrganization(from *organization) *scm.Organization {
 		Avatar: fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32/", from.Login),
 	}
 }
+
+func convertWorkspace(workspace *workspace) *scm.Organization {
+	avatar := ""
+	if workspace.Links.Avatar.Href != "" {
+		avatar = workspace.Links.Avatar.Href
+	} else {
+		avatar = fmt.Sprintf("https://bitbucket.org/account/%s/avatar/32", workspace.Slug)
```
bhumijgupta: The avatar is same for org and workspace. Can be converted to a constant

scm/driver/bitbucket/util.go
```diff
@@ -149,3 +151,122 @@ func copyPagination(from pagination, to *scm.Response) error {
...
+	} `json:"links"`
+}
+
+// fetchAllWorkspaces fetches all workspaces for the authenticated user
+func (c *wrapper) fetchAllWorkspaces(ctx context.Context) ([]string, error) {
+	var workspaceSlugs []string
+	page := 1
+	pageLen := 100
+	for {
+		path := fmt.Sprintf("2.0/user/workspaces?page=%d&pageLen=%d", page, pageLen)
```
bhumijgupta:
```suggestion
		path := fmt.Sprintf("2.0/user/workspaces?page=%d&pagelen=%d", page, pageLen)
```
https://developer.atlassian.com/cloud/bitbucket/rest/intro/#pagination

scm/driver/bitbucket/repo.go
```diff
@@ -87,34 +88,64 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
+	workspaces, err := s.client.fetchAllWorkspaces(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, ws := range workspaces {
+		perm, res, err := s.fetchRepoPerms(ctx, ws, repo)
+		if err == nil && (perm.Pull || perm.Push || perm.Admin) {
+			return perm, res, nil
+		}
```
bhumijgupta: if err != nil, we need to return the error

bhumijgupta, line 153: Need to add some comments in this code. The functionality is not clear

scm/driver/bitbucket/repo_test.go
```diff
@@ -136,17 +176,128 @@ func TestRepositoryList(t *testing.T) {
...
+	}
+}
+
+func TestRepositoryList(t *testing.T) {
+	defer gock.Off()
+
+	gock.New("https://api.bitbucket.org").
+		Get("/2.0/user/workspaces").
+		MatchParam("page", "1").
+		MatchParam("pageLen", "100").
```
bhumijgupta:
```suggestion
		MatchParam("pagelen", "100").
```
similarly in other places as well

scm/driver/bitbucket/repo.go
```diff
@@ -87,34 +88,64 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
+		copyPagination(repoStruct.pagination, response)
+		return convertRepositoryList(repoStruct), response, err
 	}
-	out := new(repositories)
-	res, err := s.client.do(ctx, "GET", path, nil, &out)
-	copyPagination(out.pagination, res)
-	return convertRepositoryList(out), res, err
+
+	return s.client.fetchReposFromAllWorkspaces(ctx, encodeListRoleOptions(opts))
```
bhumijgupta: This needs to be thought about more in depth. Previously, List(ctx, ListOptions{Page: 2, Size: 50}) did a single call to 2.0/repositories?page=2&pagelen=50, i.e. “global” page 2 across all repos

Now, the same options are passed into fetchReposFromAllWorkspaces, which calls 2.0/repositories/{workspace}?page=2&pagelen=50&role=member per workspace. So you get “page 2” for each workspace and then concatenate. That’s a different semantics (no single global page 2).

This needs to be considered

bhumijgupta (changes requested):

scm/driver/bitbucket/org.go
ritek01: will update

scm/driver/bitbucket/org.go
ritek01: cool , will do that

scm/driver/bitbucket/util.go
ritek01: sure , updated

scm/driver/bitbucket/repo_test.go
ritek01: updated in test also

scm/driver/bitbucket/repo.go
ritek01: updated to return on error

ritek01: sure will update the code with comments

scm/driver/bitbucket/repo.go
ritek01: I guess we can continue for now with current code , as mostly user would(assumption) have only access to one workspace.
And we can work on this later to fix this pagination issue.

For surely we have to fix this issue.

scm/driver/bitbucket/testdata/workspace_repo_perms.json.golden
```diff
@@ -0,0 +1,5 @@
+{
+  "Pull": true,
+  "Push": true,
+  "Admin": true
+}
```
adivishy1, line 5: I am not able to understand the parsing from workspace_repo_perms.json to workspace_repo_perms.json.golden,

what are we doing here?

scm/driver/bitbucket/testdata/workspace_repo_perms_read.json
```diff
@@ -0,0 +1,8 @@
+{
+  "values": [
+    {
+      "type": "repository_permission",
+      "permission": "read"
+    }
+  ]
+}
```
adivishy1, line 8: if we have added workspace_repo_perms_read.json, shouldnt we have workspace_repo_perms_read.json.golden and parse the response?

scm/driver/bitbucket/org.go
```diff
@@ -11,6 +11,10 @@ import (
 	"github.com/drone/go-scm/scm"
 )
 
+const (
+	avatarURLTemplate = "https://bitbucket.org/account/%s/avatar/32/"
```
adivishy1, line 15: what is 32 here?

Image
do we really need it?

adivishy1: I see its a mandatory thing to pass

Image
in that case as we dont care about the pixel size, i think we can do 16

scm/driver/bitbucket/util.go
```diff
@@ -149,3 +151,122 @@ func copyPagination(from pagination, to *scm.Response) error {
...
+			path = out.Next
+		}
+	}
+
+	return allRepos, lastRes, nil
+}
+
+// extractWorkspaceFromURL attempts to extract workspace from the client's BaseURL path.
+func (c *wrapper) extractWorkspaceFromURL() string {
+	path := strings.Trim(c.BaseURL.Path, "/")
```
adivishy1, line 482: we need some sort of try catch here

while parsing the workspaceFromURL throws some error while doing the string parsing, we need to catch it and log it and return empty

scm/driver/bitbucket/repo.go
```diff
@@ -87,34 +88,75 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
+
+	// If workspace is available, use it to fetch permissions
+	if workspace != "" {
+		repoSlug := repo
+		// Handle "workspace/repo" format by extracting just the repo slug
+		if strings.Contains(repo, "/") {
+			_, repoSlug = scm.Split(repo)
+		}
+		return s.fetchRepoPerms(ctx, workspace, repoSlug)
+	}
```
adivishy1, line 119: this block should be moved down,
if workspaces is empty, we do all the processing to get the workspaces and then it should be a common code to fetch perms

adivishy1: now we have multiple places doing the same fetchRepoPerms

scm/driver/bitbucket/repo.go
```diff
@@ -87,34 +88,75 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
+
+	// If repo is in "workspace/repo" format, split and use both parts
+	if strings.Contains(repo, "/") {
+		ws, repoSlug := scm.Split(repo)
+		return s.fetchRepoPerms(ctx, ws, repoSlug)
+	}
```
adivishy1: is this for the repo type connectors? if yes, wouldnt we have extracted the workpace from line 92 `workspace := s.client.extractWorkspaceFromURL()`

scm/driver/bitbucket/repo.go
```diff
@@ -87,34 +88,75 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
 
 // FindPerms returns the repository permissions.
 func (s *repositoryService) FindPerms(ctx context.Context, repo string) (*scm.Perm, *scm.Response, error) {
```
adivishy1, line 90: if the user provides us only repo without the workspace, we are doing the following in this impl:
1. we fetch all the workspaces the user has access to
2. iterate over each workspace and with this workspace + repo combination, we fetch the perms

now if the same repo is there in multiple workspaces, we will never know which workspace+repo the user wants to check perms for

shouldnt we make workspace a mandatory param to this endpoint? this will be a breaking change, we need to think about out impl

scm/driver/bitbucket/repo.go
```diff
@@ -88,33 +88,44 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
 	for _, ws := range workspaces {
 		perm, res, err := s.fetchRepoPerms(ctx, ws, repo)
-		if err == nil && (perm.Pull || perm.Push || perm.Admin) {
+		if err != nil {
+			return nil, res, err
+		}
+		// Return permissions if user has any access to the repository
+		if perm.Pull || perm.Push || perm.Admin {
 			return perm, res, nil
 		}
```
bhumijgupta: In case the repo is not found in the workspace, in case of a mistype, we do not return an error and the response is also nil. We should consider sending some error in that case

scm/driver/bitbucket/repo.go
```diff
@@ -88,33 +88,44 @@ func (s *repositoryService) FindHook(ctx context.Context, repo string, id string
...
 		return nil, nil, err
 	}
 
+	// Iterate through workspaces to find the repository with permissions
 	for _, ws := range workspaces {
 		perm, res, err := s.fetchRepoPerms(ctx, ws, repo)
-		if err == nil && (perm.Pull || perm.Push || perm.Admin) {
+		if err != nil {
+			return nil, res, err
+		}
```
bhumijgupta: In case the repo is not present in the workspace, would the `fetchRepoPerms` return an error, in which case, the code will fail early and not iterate through other workspaces?

scm/driver/bitbucket/repo.go
ritek01: sure , will update the handling

scm/driver/bitbucket/repo.go
ritek01: sure , will update the handling

ritek01:
Size : 100 (default max)
image
Page : 2 , Size : 2
image
Page : 3 , Size : 2
image
Page : 4 , Size : 2
image

scm/driver/bitbucket/util.go
```diff
@@ -149,3 +151,356 @@ func copyPagination(from pagination, to *scm.Response) error {
...
+	pagelen := extractPagelen(queryParams)
+	bbPage := (offset / pagelen) + 1
+	skip := offset % pagelen
+
+	var result []*scm.Repository
+	collected := 0
+	isFirstPage := true
+
+	for collected < limit {
+		repos, nextURL, err := c.fetchWorkspacePage(ctx, workspaceSlug, queryParams, bbPage)
```
bhumijgupta:
```suggestion
		repos, nextURL, err := c.fetchReposPageFromWorkspace(ctx, workspaceSlug, queryParams, bbPage)
```

scm/driver/bitbucket/util.go
```diff
@@ -149,3 +151,356 @@ func copyPagination(from pagination, to *scm.Response) error {
...
+		if err != nil {
+			return result, err
+		}
+
+		if len(repos) == 0 {
+			break
+		}
+
+		start, shouldContinue := c.getSliceStartAndCheck(repos, skip, isFirstPage)
```
bhumijgupta:
```suggestion
		start, shouldSkipPage := c.getSliceStartIdxIfWithinLimits(repos, skip, isFirstPage)
```
or something similar

scm/driver/bitbucket/util.go
```diff
@@ -149,3 +151,356 @@ func copyPagination(from pagination, to *scm.Response) error {
...
+		if shouldContinue {
+			bbPage++
+			isFirstPage = false
+			continue
+		}
+
+		end := c.getSliceEnd(len(repos), start, limit, collected)
```
bhumijgupta:
```suggestion
		end := c.getSliceEndIdx(len(repos), start, limit, collected)
```

bhumijgupta (changes requested):

scm/driver/bitbucket/util.go
ritek01: sure will update

scm/driver/bitbucket/util.go
ritek01: sure will update

scm/driver/bitbucket/util.go
ritek01: sure will update

bhumijgupta (approved):
LGTM 🚀 
Thanks for the quick work

scm/driver/bitbucket/testdata/workspace_repo_perms.json.golden
ritek01: `workspace_repo_perms.json` and `workspace_repo_perms.json.golden` are testing two different stages, not a direct file-to-file transform.

- `workspace_repo_perms.json` is the mocked Bitbucket API response for  
  `GET /2.0/workspaces/{workspace}/permissions/repositories/{repo}`.
- In `FindPerms`, that JSON is unmarshaled into `workspaceRepoPerms` (we only read `values[].permission`).
- Then `convertWorkspaceRepoPerms(...)` maps the Bitbucket permission string to `scm.Perm`:
  - `admin` -> `Pull=true, Push=true, Admin=true`
  - `write` -> `Pull=true, Push=true, Admin=false`
  - `read` (default branch) -> `Pull=true, Push=false, Admin=false`
- `workspace_repo_perms.json.golden` is the expected post-conversion internal model (`scm.Perm`) used for assertion via `cmp.Diff`.

scm/driver/bitbucket/org.go
ritek01: Yes its required for `scm.Organization` struct to have , for `32` I checked in repo we were already using `32` as size , so i added the same.

scm/driver/bitbucket/testdata/workspace_repo_perms_read.json
ritek01: this is currently not used in test , will add more test to use this file , this was added in bulk in initial commit.

scm/driver/bitbucket/repo.go
ritek01: No , its extracting the repo slug/name from the url.

scm/driver/bitbucket/util.go
ritek01: - `extractWorkspaceFromURL` does not do error-producing parsing; those ops won’t throw parse errors.
- Go has no try/catch; idiomatic handling is explicit checks.

scm/driver/bitbucket/repo.go
ritek01: already extarcted the code and moved down

scm/driver/bitbucket/repo.go
ritek01: Great callout — the ambiguity is real when only `repo` slug is provided and that slug exists in multiple workspaces.

A couple of constraints here:

- `go-scm` interface currently defines `FindPerms(context.Context, string)` (single repo identifier), so making workspace mandatory in this method would be a breaking change across providers/consumers.
- Current behavior is:
  1. prefer explicit workspace (from `workspace/repo` input or client base URL),
  2. otherwise fallback to workspace scan and return the first match.

So I agree with the concern, but I’d avoid a hard break in this PR. Suggested path:

- Keep current contract for backward compatibility.
- Document that slug-only lookup is best-effort and can be ambiguous.
- Encourage callers to pass `workspace/repo` (or base URL with workspace) for deterministic results.
- As follow-up, we can introduce a non-breaking explicit API (workspace + repo) and migrate callers to it; later deprecate slug-only fallback.

PR #352, opened 2026-02-17
scm/driver/bitbucket: fix multiple dropped errors

alrs:
This picks up several dropped `err` variables in the bitbucket package. Unit tests passed both before and after this fix.


ritek01 (approved):
LGTM ✅

